Issue #2, opened 2015-04-25
Complete move to pytest-dev


nicoddemus:
Hi @Tinche,

I have set you up as owner of this repository, and a collaborator of @pytest-dev, so I think all is good. :smile:

To complete the transition, here are a couple of things it would be nice to do:

- [ ] Update links in code (`setup.py`, `README`, etc) and documentation;
- [ ] Update Travis connection;
- [ ] Update coveralls connection;

Feel free to ask for help with any of these, here or in the mailing list.

Welcome! :smile:

GMLudo:
Congratulations @Tinche and @nicoddemus, it's good to see that this project becomes an official repository of `pytest` ;-)

PR #3, opened 2015-04-25
Transition to pytest-dev.

Tinche:
Fixes for #2.

Getting some coverage failures (submitting returns 500), will look into it at a later time.
Also will cut a new release (0.1.3) once this is merged, to update the URL on PyPI.


nicoddemus:
Good job! :smile:

PR #8, opened 2015-06-05
Various simplifications

Lucretiel:
Removed unneeded local finalizer in event_loop
Simplified code in unused_tcp_port


Tinche:
Looks good to me, thanks!

PR #9, opened 2015-07-17
Get rid of using external resource in tests

asvetlov:
Fetching data from http://httpbin.org makes tests execution slow.
What we really need is checking for asyncio functionality, not HTTP page fetching.

### DIFF
--- a/tests/test_simple.py
+++ b/tests/test_simple.py
@@ -6,36 +6,25 @@ import pytest
 
 
 @asyncio.coroutine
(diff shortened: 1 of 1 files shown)


Tinche:
Alright, the changes LGTM, the coverage LGTM. Thanks for the refactor!

asvetlov:
That's only beginning, I have a couple of more important refactorings.
Will introduce these soon.

PR #14, opened 2015-08-01
Unused TCP port factory fixture. Bump to 0.2.0.

Tinche:
I've decided to simplify the new fixture a little; now it's a callable which returns a different port each invocation.

@nicoddemus Could you please do a small review before merging?


pytest_asyncio/plugin.py
```diff
@@ -88,3 +88,17 @@ def unused_tcp_port():
...
+
+@pytest.fixture
+def unused_tcp_port_factory():
+    """A factory function, producing different unused TCP ports."""
+    produced = set()
+
+    def factory():
+        port = unused_tcp_port()
+        while port in produced:
+            port = unused_tcp_port
```
nicoddemus: I think this should be `port = unused_tcp_port()`.

nicoddemus:
Hmm coverage has decresed by 2%, but I can't access the source when I try to access coveralls...

pytest_asyncio/plugin.py
Tinche: Good catch! Of course the bug is in the section not tested. I'll add a test and get coverage up to 100%.

nicoddemus:
Looks good to me, nice work! :smile:

nicoddemus:
(Not sure if I was supposed to merge it)

PR #16, opened 2015-08-25
trollius support

chassing:
use trollius instead asyncio for python 2

### DIFF
--- a/pytest_asyncio/plugin.py
+++ b/pytest_asyncio/plugin.py
@@ -1,4 +1,9 @@
-import asyncio
+try:
+    # python 3
+    import asyncio
+except ImportError:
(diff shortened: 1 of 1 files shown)


Tinche:
I'm not really sold on the idea of supporting Trollius/Python 2 in pytest-asyncio. Is this something you personally need?

chassing:
yes indeed. I've a project which supports python 2 and 3 with trollius and asyncio

chassing:
any news about it?

Tinche:
I've thought about it, but since the Trollius project is now deprecated (http://trollius.readthedocs.org/deprecated.html#deprecated), I don't think it's worth it. Sorry.

PR #17, opened 2015-12-17
Support coroutines with async and await syntax

dirn:
PEP 492 added support for defining coroutines using `async def` rather
than having to decorate the function with `@asyncio.coroutine`. The new
`_is_coroutine` function will match any coroutines created with the
decorator using `inspect.isgeneratorfunction` as well as those created
through the async and await syntax using `asyncio.iscoroutinefunction`.

The tests for this need to be added in an unconventional way. `async`
and `await` cause syntax errors in versions of Python prior to 3.5.
Rather than causing the tests for 3.3 and 3.4 to fail, the tests are
being defined as a string and then compiled and executed into the
module. If a time ever comes that this library no longer supports
versions prior to 3.5, this new module can be removed entirely.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,10 @@
 language: python
-python: 3.3
+python: 3.5
 
 env:
     - TOX_ENV=py33
     - TOX_ENV=py34
+    - TOX_ENV=py35
 
 install:
     - pip install tox
--- a/pytest_asyncio/plugin.py
+++ b/pytest_asyncio/plugin.py
@@ -6,6 +6,11 @@ import socket
 import pytest
 
 
+def _is_coroutine(obj):
+    """Check to see if an object is really an asyncio coroutine."""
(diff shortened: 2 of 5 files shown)


Tinche:
Thanks for your contribution and working to make pytest-asyncio better!

I'll do my best to take a look at this tomorrow and merge it in.

Tinche:
Alright, I think this looks good.

There's a better way of handling files with 3.5 only syntax: customize the test collector a little. I have this done and will apply a small commit on top of your work.

After that, I'll update the README, change the examples to mainly use the new async/await syntax, bump the version up and do a release!

Thanks for your work! Do you have a Twitter handle for me to mention you when I announce the release?

dirn:
I was looking for a better way to do it but couldn't seem to get the collector to work right. I'm glad you know how to do that. 

I'm @dirn on Twitter. Thanks.

PR #22, opened 2016-05-13
gracefully handle contexts with no existing event loop

csboling:
On Windows 7 with Python 3.5.1, running the test suite produced several occurrences of the following exception:

```
_____________ ERROR at setup of test_unused_port_factory_fixture ______________

request = <SubRequest 'event_loop' for <Function 'test_unused_port_factory_fixture'>>

    @pytest.fixture
    def event_loop(request):

>       policy.get_event_loop().close()

..\pytest_asyncio\plugin.py:72:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <asyncio.windows_events._WindowsDefaultEventLoopPolicy object at 0x0000000003D40860>

    def get_event_loop(self):
        """Get the event loop.

            This may be None or an instance of EventLoop.
            """
        if (self._local._loop is None and
            not self._local._set_called and
            isinstance(threading.current_thread(), threading._MainThread)):
            self.set_event_loop(self.new_event_loop())
        if self._local._loop is None:
            raise RuntimeError('There is no current event loop in thread %r.'
>                              % threading.current_thread().name)
E           RuntimeError: There is no current event loop in thread 'MainThread'.`
```

By checking if `policy.get_event_loop()` raises a `RuntimeError` before attempting to close the existing event loop in the `event_loop` fixture, this error is avoided.

### DIFF
--- a/pytest_asyncio/plugin.py
+++ b/pytest_asyncio/plugin.py
@@ -71,7 +71,7 @@ def event_loop(request):
 
     try:
         current_loop = policy.get_event_loop()
-    except RuntimeError:
+    except (RuntimeError, AssertionError):
         pass
     else:
         current_loop.close()
--- a/pytest_asyncio/plugin.py
+++ b/pytest_asyncio/plugin.py
@@ -69,13 +69,18 @@ def event_loop(request):
     """Create an instance of the default event loop for each test case."""
     policy = asyncio.get_event_loop_policy()
 
-    policy.get_event_loop().close()
+    try:
+        current_loop = policy.get_event_loop()
+    except RuntimeError:
+        pass
+    else:
+        current_loop.close()
+    finally:
+        event_loop = policy.new_event_loop()
+        policy.set_event_loop(event_loop)
 
-    event_loop = policy.new_event_loop()
-    policy.set_event_loop(event_loop)
-
-    request.addfinalizer(event_loop.close)
-    return event_loop
+        request.addfinalizer(event_loop.close)
+        return event_loop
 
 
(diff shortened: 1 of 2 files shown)


csboling:
The test (`test_asyncio_marker_with_non_global_loop`) was also failing on Python 3.4 because `get_event_loop` used to raise `AssertionError` rather than `RuntimeError` if no event loop exists. This behavior was changed in [this](https://github.com/python/cpython/commit/dff40045482df29404fc9478c0f1a9ed310cfb24) commit. This has been resolved in the `event_loop` fixture by catching `AssertionError` as well, though I'm not sure I like this solution. The implementation of `get_event_loop` in python/asyncio also raises a RuntimeError, so this only affects version 3.4. This PR now also splits the test case in question depending on the Python version.

Tinche:
Hello,

thanks for putting effort into this! I'll try looking over this over the weekend.

Tinche:
Hi,

sorry for the delay in replying, I was busy IRL. I recently merged in some changes aimed at making overriding the event_loop fixture easier (this affects the `@pytest.mark.asyncio` marker as well). Anyway, the setting of the event loop is handled differently now (we don't fetch the existing loop at the start and close it, just create a new one and close that at the end). 

Could you check if these changes work for you?

If not, could you please rebase?

csboling:
Yep, tests now pass on my system and this solution is much more elegant. Closing.

PR #23, opened 2016-05-16
Feature/pluggable event loops

Tinche:


Juanlu001:
I just wanted to say that this change makes the fixture unusable in Python < 3.5.3 because of this bug:

https://bugs.python.org/issue28613

Hope it helps someone!

Issue #26, opened 2016-06-01
Tests don't fail anymore


a7p:
having a minimal:
```
from pytest import mark
@mark.xfail(reason="I need a failing 
def test_fail_please():
     assert False

@mark.xfail(reason="I'm testing for a pytest-asyncio-bug - if I don't fail something is broken.")
@mark.asyncio
def test_fail_async():
     assert False
```

`test_fail_async` fails as expected with asyncio==0.3.0, but succeeds with asyncio==0.4.0.

(running python3.5.1)

Tinche:
Hi, thanks for reporting, it's a known issue: #25. Expect a new release later today :)

a7p:
arg - the title did not make me look at the issue ;).

Tinche:
Hi, I've just pushed 0.4.1 to PyPI. Can you confirm it fixes the issue for you?

a7p:
thank you very much - it does.

PR #27, opened 2016-06-01
adds a test to reproduce issue #26

a7p:
minimal test to reproduce https://github.com/pytest-dev/pytest-asyncio/issues/26


Tinche:
This is a useful test to have, thanks! I'll just add `strict=True`.

a7p:
ah, cool, thx too, did not know about `strict`.

PR #37, opened 2016-11-27
Switch to setuptools-scm for versioning

hayderimran7:
Following other pytest projects like pytest-html,
that use setuptool-scm to manage version 
https://github.com/pytest-dev/pytest-html/commit/d9c0b8782b44df9250305988e76d095fe75e6e15 .  
It would be nice other projects follow the suit and
let setuptools-scm manage version of this project.


Tinche:
Hi!

This looks interesting. I'm generally in favor, but I need to get some free time to read up on this and play around with it. :)

nicoddemus:
For the record it works great, and will also solve #92 because `setuptools_scm` will add all versioned files to the package automatically.

nicoddemus:
I've fixed the conflicts and adopted the strategy of writing a `_version.py` file so we can keep having the `__version__` attribute in the package. 😉

nicoddemus:
This also paves the way to automatic deployments using Travis, which are highly recommended. Let me know if I can help @Tinche!

blueyed:
FWIW I think https://github.com/zsimic/setupmeta is even nicer.

seifertm:
Rebased the PR to master and added `setuptools_scm` as dependency to the CI steps.

I also cloned my local `pytest-asyncio` git repo and and ran `tox` in a fresh virtualenv. Tox successfully picked up the dependency on `setuptools_scm`.

.github/workflows/main.yml
```diff
@@ -57,7 +57,7 @@ jobs:
           python-version: "3.9"
 
       - name: "Install poetry, check-wheel-content, and twine"
-        run: "python -m pip install wheel twine check-wheel-contents"
+        run: "python -m pip install check-wheel-contents setuptools_scm twine wheel"
```
asvetlov, line 60:
```suggestion
        run: "python -m pip install build check-wheel-contents twine"
```

.github/workflows/main.yml
```diff
@@ -57,7 +57,7 @@ jobs:
...
       - name: "Build package"
         run: "python setup.py sdist bdist_wheel"
```
asvetlov, line 62:
```suggestion
        run: "python -m build"
```
The officially recommended approach that builds a tarball, then creates a wheel from the tarball.
It avoids to make a silly error when a file required by wheel is not included in the tarball.

setup.cfg
```diff
@@ -1,6 +1,5 @@
 [metadata]
 name = pytest-asyncio
-version = attr: pytest_asyncio.__version__
```
asvetlov, line 3: The version is substituted by SCM

asvetlov (approved):
Looks good, please apply my tiny improvements before pressing the 'squash and merge' button.

setup.cfg
seifertm: Thanks, I re-added the line.

seifertm:
Without the `pip install setuptools_scm` in the testing job, `tox` didn't pick up the dependency on `setuptools_scm` correctly and the test run failed due to a missing `_version.py` file. After adding `setuptools_scm` to `setup_requires` in `setup.cfg` everything worked. This is also how pytest-xdist handles it. They define the dependency in `setup.cfg`, [1] but configure `setuptools_scm` in `pyproject.toml` [2].

[1] https://github.com/pytest-dev/pytest-xdist/blob/a1a6c77028d1c2215f2457482602d214f7f9f532/setup.cfg#L43
[2] https://github.com/pytest-dev/pytest-xdist/blob/a1a6c77028d1c2215f2457482602d214f7f9f532/pyproject.toml#L10-L11

asvetlov:
Interesting. I thought that the single source of `setuptools_scm` dependency (`pyproject.toml`) should be enough.
Can live with this sort of duplication, sure!
Thanks for the double-checking.

PR #45, opened 2017-03-11
Add async fixture support (#41)

### LINKED ISSUE #41: async generators pytest fixtures
Python 3.6 have introduced new async generator entity:
```python
async def foo():
    yield 'Foo'
```

When trying to create yield fixture in pytest (or a simple fixture but with `async def`) and decorate it  with `@pytest.mark.asyncio` i receive a `coroutine` instead of fixture result. 

Is there a way to make `pytest.mark.asyncio` recognize fixtures as coroutine functions or async generator functions and behave properly on them?

sashgorokhov:
* Added async generator fixture support
* Added async fixture support

There are some problems I didnt solved yet:

* `@asyncio.coroutine` fixtures. From inside the code, such fixtures appears as common generators - didnt realized how to distinguish them from, for example, user-defined fixtures that supposed to return generators.
* Async fixtures scope. Currently only `function` scope is supported. I'll try to solve it. The main problem that async fixtures must (or not?) be run in the same event loop as test function. With `module` and `session` scoped fixtures this could be a problem.

Closes #41


Tinche:
Honestly this looks like great work! Could you please add a little documentation to the readme?

sashgorokhov:
@Tinche yes I'll add documentation for these async fixtures this week. What about async fixtures `function`-only scope? Is that ok?

Tinche:
Regarding function-scope only fixtures, I'll take what I can get. :)

Tinche:
Thank you for your contribution!

Issue #55, opened 2017-06-04
AttributeError: module 'asyncio' has no attribute 'ensure_future' with pytest-xdist


orf:
If you attempt to use `pytest-asyncio` with `pytest-xdist`, and run `py.test -n 4` it will fail with the exception in the title [at this line](https://github.com/pytest-dev/pytest-asyncio/blob/master/pytest_asyncio/plugin.py#L155).

Running it without `-n 4` works as expected. I'm not sure what's going on here or why that attribute would be missing.

orf:
Hmm, that MR seems to pass. On my mac, I get:

```
⇒  py.test -n 4 .                                                                                                          (env: x)
=========================================================== test session starts ============================================================
platform darwin -- Python 3.6.1, pytest-3.1.1, py-1.4.33, pluggy-0.4.0
rootdir: /Users/x/PycharmProjects/x, inifile:
plugins: xdist-1.16.0, asyncio-0.6.0
gw0 [2] / gw1 [2] / gw2 [2] / gw3 [2]
scheduling tests via LoadScheduling
.F
================================================================= FAILURES =================================================================
___________________________________________________________ test_basic_pipeline ____________________________________________________________
[gw0] darwin -- Python 3.6.1 /Users/x/.virtualenvs/x/bin/python3.6
pyfuncitem = <Function 'test_basic_pipeline'>

    @pytest.mark.tryfirst
    def pytest_pyfunc_call(pyfuncitem):
        """
        Run asyncio marked test functions in an event loop instead of a normal
        function call.
        """
        for marker_name, fixture_name in _markers_2_fixtures.items():
            if marker_name in pyfuncitem.keywords:
                event_loop = pyfuncitem.funcargs[fixture_name]

                funcargs = pyfuncitem.funcargs
                testargs = {arg: funcargs[arg]
                            for arg in pyfuncitem._fixtureinfo.argnames}
                event_loop.run_until_complete(
>                   asyncio.ensure_future(
                        pyfuncitem.obj(**testargs), loop=event_loop))
E               AttributeError: module 'asyncio' has no attribute 'ensure_future'

../../.virtualenvs/x/lib/python3.6/site-packages/pytest_asyncio/plugin.py:154: AttributeError
==================================================== 1 failed, 1 passed in 0.59 seconds ====================================================
x@mac:~/PycharmProjects/x|master⚡
⇒
```

But also:

```
⇒  py.test                                                                                                                 (env: x)
[...]

PR #56, opened 2017-06-04
Test with pytest-xdist

orf:
Re #55

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -4,7 +4,8 @@ minversion = 2.5.0
 
 [testenv]
 extras = testing
-commands = coverage run -m pytest {posargs}
+deps = pytest-xdist
+commands = coverage run -m pytest -n auto {posargs}
 
 [testenv:coverage-report]
 deps = coverage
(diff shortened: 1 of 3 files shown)


tox.ini
```diff
@@ -6,6 +6,6 @@ deps =
     pip >= 6
     -rtest_requirements.txt
 commands =
-    coverage run --source pytest_asyncio -m py.test
+    coverage run --source pytest_asyncio -m py.test -n 4
```
discosultan: Why not use `-n auto`? It is described [here](https://github.com/pytest-dev/pytest-xdist#speed-up-test-runs-by-sending-tests-to-multiple-cpus) and imo should be the default choice unless there's a reason to fix the parallelism.

discosultan: Good to know about the Travis issue. Glad there's a solution 🙂

orf:
I've updated the PR to run xdist on travis, plus made it run tests on MacOS.

nicoddemus:
@orf unfortunately `pytest 3.7.0` is incompatible with `pytest-asyncio`: https://github.com/pytest-dev/pytest-asyncio/issues/89.

It should be fixed in `pytest 3.7.1` though (https://github.com/pytest-dev/pytest/pull/3754), so I suggest to pin to `pytest!=3.7.0` for now.

PR #60, opened 2017-07-05
[Documentation] Finalizers and async fixtures.

### LINKED ISSUE #59: Add async finalizers to the request-context
Hi,
My project is Python 3.5 only, and it's therefore impossible for me to use async generators in fixtures. But we might be able to workaround this if it were possible to define async finalizers to `request`.

E.g.
```python
@pytest.fixture
async def foo_fixture(request):
    async def fin():
        await some_stuff()
    request.addasyncfinalizer(fin)
    return 42
```
Then it would await for each async finalizers before tear down.

Not sure whether it would involve monkey patching on the request object, or what would be acceptable. Maybe there are also more proper solutions that doesn't involve `request` or that make use of already existing features? I already tried to use `loop.create_task()` in a standard finalizer, unfortunately, the created task seem to execute after teardown which produces the following error message:
`2017-06-26 18:15:50,262 - asyncio - ERROR - Task was destroyed but it is pending!`

Many thanks!

gilbsgilbs:
Fixes #59 

Thanks :) .

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -130,6 +130,21 @@ to redefine the ``event_loop`` fixture to have the same or broader scope.
 Async fixtures need the event loop, and so must have the same or narrower scope
 than the ``event_loop`` fixture.
 
+Python 3.5 lacks support for async generators. If you really need support for
+Python 3.5, you can still mimic a tear down using pytest's finalizers. Notice
+that the event loop is not running inside finalizers:
+
+.. code-block:: python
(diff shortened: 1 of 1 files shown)


agronholm:
I think #62 invalidates this PR.

gilbsgilbs:
@agronholm Totally, and it also closes #59.

PR #63, opened 2017-07-18
Packaging and testing cleanups

agronholm:
Mostly cosmetic simplifications and improvements in the testing and packaging setup.
Highlights:

* Use pytest-cov for test coverage
* Removed test_requirements.txt in favor of a `testing` extras (`pip install -e .[testing]`)
* Added `python_requires` to packaging metadata
* Use Pathlib for reading file contents

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,10 +3,8 @@ python:
   - "3.5"
   - "3.6"
 
-install:
-    - pip install tox tox-travis
+install: pip install tox-travis coveralls
 
 script: tox
 
-after_success:
-    - pip install coveralls && cd tests && coveralls
+after_success: coveralls
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,3 +1,13 @@
+[coverage:run]
+source = pytest_asyncio
+
+[coverage:report]
+show_missing = true
+
+[tool:pytest]
+addopts = -rsx --tb=short --cov
+testpaths = tests
+
 [metadata]
 # ensure LICENSE is included in wheel metadata
 license_file = LICENSE
--- a/setup.py
+++ b/setup.py
@@ -1,19 +1,11 @@
-import codecs
-import os
 import re
+from pathlib import Path
+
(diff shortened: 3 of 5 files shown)


Tinche:
For the failing test, I suggest making tests/async_fixtures/test_async_gen_fixtures_35.py collectable only for 3.5 (the refactor removed async_generator as a test dependency, which I think is the right move).

agronholm:
No, the accidental removal of async_generator as testing dependency was a mishap. I will add it back. Such generators must be usable from Python 3.6+ and should also be tested that way.

Tinche:
I'd say 3.6+ should just use normal async generators, no?

agronholm:
And if you're running 3.5 compatible code on 3.6?

agronholm:
Let me rephrase: test code written for 3.5 should be runnable on 3.6 too, and the inclusion of `async_generator` in the test requirements ensures that the test suite covers this as well.

Tinche:
Right, you have a point. I don't see the harm including it in the test requirements in any case.

Tinche:
One last thing: I see you removed the metadata section from setup.cfg. Is the license file still being included in the wheel? I had an explicit request for this from a downstream repackager.

agronholm:
I see no references to such functionality in setuptools or wheel. Nor have I seen this anywhere else. Can you point me to documentation that says this is a thing?

Tinche:
Right, it's kinda obscure. Here's a link, search for 'metadata' inside: https://hynek.me/articles/sharing-your-labor-of-love-pypi-quick-and-dirty/

You can test it out yourself, wheels are zips. The license should be in there after `python setup.py bdist_wheel`.

agronholm:
Seems like an undocumented option. I'll put it back.

agronholm:
Note that the official wheel documentation says the following:
> Older versions of bdist_wheel supported passing requirements in a now-deprecated [metadata] section in setup.cfg.

Tinche:
Thanks!

Issue #52, opened 2017-05-29
Latest version causing test failures


vangheem:
Getting:

```
E           RuntimeError: Task <Task pending coro=<pytest_fixture_setup.<locals>.wrapper.<locals>.setup() running at /home/travis/build/plone/guillotina/eggs/pytest_asyncio-0.6.0-py3.6.egg/pytest_asyncio/plugin.py:120> cb=[_run_until_complete_cb() at /opt/python/3.6-dev/lib/python3.6/asyncio/base_events.py:176]> got Future <_GatheringFuture pending> attached to a different loop
```

on many of my tests now.

Downgrading to 0.5.0 works. Is there something I'm doing wrong in my tests?

Tinche:
Not necessarily, might be a bug in our fixture handling code.

Can you get me a minimal example?

vangheem:
I'll see if I have time to do that later. The package the test failures started in is open source though. 

Here is failure output: https://travis-ci.org/plone/guillotina/builds/237114876

and here is the test server being created: https://github.com/plone/guillotina/blob/master/guillotina/tests/conftest.py#L198

But that's a lot of code to digest and not sure there is anything obvious there.

I'll see what I can do when I have some more time.

Tinche:
As far as I can see, your tests are actually pytest-aiohttp, not pytest-asyncio tests. You have async def tests without the `@pytest.mark.asyncio` marker. 

I see you have some async def fixtures too, which weren't supported in pytest-asyncio until 0.6. So these were being picked up by pytest-aiohttp too.

So I'm guessing now pytest-asyncio and pytest-aiohttp are in conflict.

spumer:
I have failure on 0.6.0 too, but due setup `event_loop` fixture.
```
fixturedef = <FixtureDef name='event_loop' scope='function' baseid='' >
request = <SubRequest 'event_loop' for <Function 'test_get_gen_products'>>
... # here a lot of code
        if fixturedef.argname == "event_loop" and 'asyncio' in request.keywords:
            loop = outcome.get_result()
            for kw in _markers_2_fixtures.keys():
                if kw not in request.keywords:
                    continue
                policy = asyncio.get_event_loop_policy()
>               old_loop = policy.get_event_loop()


    def get_event_loop(self):
        """Get the event loop.
    
            This may be None or an instance of EventLoop.
            """
        if (self._local._loop is None and
            not self._local._set_called and
[...]

PR #64, opened 2017-07-24
handle case when no event loop exists (support pytest-aiohttp)

smagafurov:
Solves this:
- https://github.com/pytest-dev/pytest-asyncio/issues/52

And this:
- https://github.com/aio-libs/aiohttp/issues/1069

Now pytest-asyncio and pytest-aiohttp can act together with this conftest.py:

```python
from aiohttp.test_utils import loop_context


@pytest.yield_fixture
def loop():
    with loop_context() as _loop:
        yield _loop


@pytest.yield_fixture
def event_loop(loop):
    yield loop
```

### DIFF
--- a/pytest_asyncio/plugin.py
+++ b/pytest_asyncio/plugin.py
@@ -129,7 +129,12 @@ def pytest_fixture_setup(fixturedef, request):
             if kw not in request.keywords:
                 continue
             policy = asyncio.get_event_loop_policy()
-            old_loop = policy.get_event_loop()
+            try:
+                old_loop = policy.get_event_loop()
+            except RuntimeError as exc:
+                if 'no current event loop' not in str(exc):
+                    raise
+                old_loop = None
             policy.set_event_loop(loop)
             fixturedef.addfinalizer(lambda: policy.set_event_loop(old_loop))
 
--- a/tests/test_simple.py
+++ b/tests/test_simple.py
@@ -7,7 +7,7 @@ import pytest_asyncio.plugin
 
 
 @asyncio.coroutine
-def async_coro(loop):
+def async_coro(loop=None):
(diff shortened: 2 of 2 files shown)


smagafurov:
@asvetlov, hi

see conftest.py in my comment above

How do you think, is it ok to run pytest-asyncio and pytest-aiohttp together in such manner?

spumer:
Yea! This is very usefull!

asvetlov:
Looks good at first glance

argaen:
I really should check open PR first before doing mine 😓 . Looks good to me too, only thing I'm missing there is a test, you can check https://github.com/pytest-dev/pytest-asyncio/pull/65, copy paste for the test should do it.

@Tinche any intentions on merging this? :)

Tinche:
The patch looks simple enough, I'd be willing to merge it in. But wouldn't a better course of action be to reach out to pytest-aiohttp and see if they would be willing to depend on pytest-asyncio, instead of merging in workarounds?

argaen:
Now it's `pytest-aiohttp` the one having this edge case but anyone can do an `asyncio.set_event_loop(None)` in a pytest fixture and then `pytest-asyncio` starts crashing. IMO this should be solved regardless of what other libraries do

smagafurov:
test added (copy pasted from #65)

bugok:
+1

I'm hitting this problem as well. Would be happy to see this getting merged in.

Tinche:
Alright, thanks for the contribution. My plea for pytest-asyncio and pytest-aiohttp to work better together still stands :)

smagafurov:
can we release minor version with this fix?

Tinche:
I'll get it out over the weekend.

PR #65, opened 2017-09-08
asyncio marker works if no event loop is set

argaen:
There is at least one edge case where there is no event loop when https://github.com/pytest-dev/pytest-asyncio/compare/master...argaen:catch_when_no_loop?expand=1#diff-4d85e8313d8e67cb777137411a11ac86R133 is reached.

When combining pytest-aiohttp and this package (only with 0.6.0 version) tests are crashing saying there is no event loop running. I believe the error is produced because pytest-aiohttp does the following: https://github.com/aio-libs/aiohttp/blob/master/aiohttp/pytest_plugin.py#L145 and then when pytest-asyncio reaches the line where it has to get the loop, there is no loop. This PR provides a try/except for controlling this case.

I was checking previous versions of this package to see why it wasn't crashing but from what I can see it has changed quite a lot

### DIFF
--- a/pytest_asyncio/plugin.py
+++ b/pytest_asyncio/plugin.py
@@ -129,9 +129,12 @@ def pytest_fixture_setup(fixturedef, request):
             if kw not in request.keywords:
                 continue
             policy = asyncio.get_event_loop_policy()
-            old_loop = policy.get_event_loop()
+            try:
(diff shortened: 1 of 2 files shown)


argaen:
Oh well, closing in favor of https://github.com/pytest-dev/pytest-asyncio/pull/64

PR #74, opened 2018-01-12
preliminary debian package work

themusicgod1:
Since pytest-asyncio is required by other software, and since there is an effort to try to get that software into Debian GNU/Linux it would be useful to have the files required to be a debian project synced with this repository. This as a side effect would also help get pytest-asyncio (and software depended upon it) into Ubuntu, Mint, Tails, Devuan, and dozens of other distributions both current and future if successful. On the Debian side, this effort is tracked in the following Request for Package (RFP): https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=886338

I have made a fork and added what seems to be the minimum necessary changes to actually be a Debian package, though since I am not actually a DM/DD/maintainer the name will eventually have to be changed by whoever with the actual credentials required to be a maintainer the Debian project chooses to be the maintainer, but it's a first attempt anyway. It does build and allows for pytest-asyncio to be installed on a Debian GNU/Linux system.


auvipy:
should we close it?

asvetlov:
I think yes, please close.
Downstream add required files on their own.
I support dozens of python libraries, very many of them are part of Debian and Fedora distros.
Nobody asked me to include the Debian package configuration.

PR #88, opened 2018-07-09
Python 3.7 support

### LINKED ISSUE #87: Stop using ProcessPoolExecutor as a default event loop executor
This usage is soon to be deprecated and prohibited by the standard library (https://bugs.python.org/issue34075), and it currently causes issues in 3.7.0 (https://bugs.python.org/issue34073). This issue was discovered when making changes in `pytest-asyncio` to make the tests work under 3.7.0. This affects the `event_loop_process_pool` fixture and the `pytest.mark.asyncio_process_pool` marker.

jonafato:
- Enable tests and declare support for Python 3.7
- Use `async` and `await` in all tests
-  Remove ProcessPoolExecutor support

   Remove the `event_loop_process_pool` fixture and the
   `pytest.mark.asyncio_process_pool` marker to address deprecation and
   removal detailed in https://bugs.python.org/issue34075.

Fixes #87.

~Note: this change may be blocked by #87. The behavior of the
`ProcessPoolExecutor` tests is somewhat flaky in 3.7 and may cause the
test suite to hang.~
Because this change is blocked by #87, I've rolled the fix into this PR.
I can split this up if desired.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,20 @@
 language: python
-python:
-  - "3.5"
-  - "3.6"
+matrix:
+  include:
+    - python: 3.5
+      env: TOX_ENV=py35
+    - python: 3.6
+      env: TOX_ENV=py36
+    - python: 3.7
+      env: TOX_ENV=py37
+      # TODO: the dist and sudo keys are currently needed to use Python 3.7.
+      # They should be removed once Travis-CI supports 3.7 on the default image.
+      dist: xenial
+      sudo: true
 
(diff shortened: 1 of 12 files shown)


jonafato:
Per my note, https://travis-ci.org/pytest-dev/pytest-asyncio/jobs/401926629 is definitely hanging.

discosultan:
Curious as to what the status is on this. Is there anything blocking that the community could chip in?

Tinche:
Thank you very much @jonafato !

PR #97, opened 2018-09-27
Fix: Avoid warning on latest Pytest versions

JoseKilo:
Pytest added a deprecation warning to signal fixtures being called
directly as functions.

Ref: https://github.com/pytest-dev/pytest/issues/3661
Ref: https://docs.pytest.org/en/latest/deprecations.html#calling-fixtures-directly

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -4,7 +4,7 @@ minversion = 2.5.0
 
 [testenv]
 extras = testing
-commands = coverage run -m pytest -W error {posargs}
+commands = coverage run -m pytest {posargs}
 
 [testenv:coverage-report]
 deps = coverage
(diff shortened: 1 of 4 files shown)


tox.ini
```diff
@@ -4,7 +4,7 @@ minversion = 2.5.0
 
 [testenv]
 extras = testing
-commands = coverage run -m pytest {posargs}
+commands = coverage run -m pytest -W error {posargs}
```
nicoddemus: Probably better to add this to `pytest.ini` instead:

```ini
[pytest]
filterwarnings = error
```

dbuse:
Anything blocking this from merging? Would be really nice to get this released!

Tinche:
Thanks!

PR #99, opened 2018-10-02
Add unused port helpers for UDP

dbuse:
Extends the unused_tcp_port and unused_tcp_port_factory mechanisms for UDP ports.


eirnym:
Could you make `socket_type` setting explicit for TCP as well? I understand that it's default, but it would be much easier to read 

Also both factories `unused_udp_port_factory` and `unused_tcp_port_factory` are essencially the same. The only difference is `socket_type`.

dbuse:
> Could you make `socket_type` setting explicit for TCP as well? I understand that it's default, but it would be much easier to read

Agreed, that makes sense.

> Also both factories `unused_udp_port_factory` and `unused_tcp_port_factory` are essencially the same. The only difference is socket_type.

True. My intention was to stay fully backward-compatible. So changing `unused_tcp_port_factory` to accept a socket type would invalidate the `tcp' in its name. But I could extract the implementation into a common function (like with the `_unused port' function) and make the fixtures a slim wrapper.

> Additionally, please, spread documentation to fixtures.

Also agreed.

I'll look into the code changes soon.

pytest_asyncio/plugin.py
```diff
@@ -193,10 +198,28 @@ def unused_tcp_port_factory():
...
+        while port in produced:
+            port = _unused_port()
+
+        produced.add(port)
+
+        return port
+    return factory
+
+
+@pytest.fixture
```
asvetlov, line 212:
```suggestion
@pytest.fixture(scope=session)
```
To reflect `unused_tcp_port_factory` fixture definition

asvetlov (approved):

PR #102, opened 2018-12-12
Support async tests which use Hypothesis

### LINKED ISSUE #93: Support for hypothesis in async methods
`hypothesis` exits early when used on pytest-asyncio tests, complaining that the test returns a coroutine instead of None. This issue talks about how trio added support for hypothesis: https://github.com/HypothesisWorks/hypothesis/issues/968

I can work around this with a decorator:

```py
def run_in_event_loop(async_func):
    @functools.wraps(async_func)
    def wrapped(operations, queue_size, add_size, get_size, event_loop):
        event_loop.run_until_complete(asyncio.ensure_future(
            async_func(operations, queue_size, add_size, get_size, event_loop),
            loop=event_loop,
        ))
    return wrapped

@hypothesis.given(...)
@run_in_event_loop
@pytest.mark.asyncio
async def test_...
```

But it looks like the recommended approach is to tinker with the hypothesis attribute added to the test, like: https://github.com/python-trio/pytest-trio/pull/44/files#diff-aacce3c1178656a02aecc59c94209433R231

Is this something pytest-asyncio is interested in supporting?

Zac-HD:
Closes #93.

### DIFF
--- a/pytest_asyncio/plugin.py
+++ b/pytest_asyncio/plugin.py
@@ -1,6 +1,7 @@
 """pytest-asyncio implementation."""
 import asyncio
 import contextlib
+import functools
 import inspect
 import socket
 
@@ -139,7 +140,8 @@ def pytest_pyfunc_call(pyfuncitem):
     function call.
     """
     for marker_name, fixture_name in _markers_2_fixtures.items():
-        if marker_name in pyfuncitem.keywords:
+        if isasyncgenfunction(pyfuncitem.obj) \
+                and marker_name in pyfuncitem.keywords:
             event_loop = pyfuncitem.funcargs[fixture_name]
 
             funcargs = pyfuncitem.funcargs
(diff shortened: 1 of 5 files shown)


pytest_asyncio/plugin.py
```diff
@@ -139,7 +140,8 @@ def pytest_pyfunc_call(pyfuncitem):
     function call.
     """
     for marker_name, fixture_name in _markers_2_fixtures.items():
-        if marker_name in pyfuncitem.keywords:
+        if inspect.isasyncgenfunction(pyfuncitem.obj) \
```
asvetlov: `pytest-asyncio` claims that the minimum supported Python is 3.5.
`isasyncgenfunction` is available in Python 3.6+ only

pytest_asyncio/plugin.py
```diff
@@ -152,11 +154,39 @@ def pytest_pyfunc_call(pyfuncitem):
             return True
 
 
+def wrap_in_sync(func):
+    """Return a sync wrapper around an async function."""
+
+    @functools.wraps(func)
+    def inner(**kwargs):
+        loop = asyncio.get_event_loop_policy().new_event_loop()
```
asvetlov: Why not bare `asyncio.new_event_loop()` here?

asvetlov: `pytest-asyncio` works slightly different with event loops:
1. It installs the loop as default
2. Restores previously used default loop after the test finishing

Does your change follow this behavior?

Zac-HD: > Why not bare `asyncio.new_event_loop()` here?

There are so many ways to get an event loop that I just copied the `event_loop` fixture; I assume there's a reason it was chosen there.

> Does your change [set and restore the default loop]?

Not yet!  One moment...

pytest_asyncio/plugin.py
Zac-HD: Fixed :smile:

Tinche:
I think this is looking good! Looking forward to merging it in, as an avid user of Hypothesis :) Thank you for your contribution.

Gonna leave it open for a few days to solicit additional comments.

How difficult would it be in the future to add support for:

* fixtures using Hypothesis
* Hypothesis state machines using async steps?

Having those as well would really rock.

Zac-HD:
Awesome, glad I could help!  Hopefully it will be released to users before the holidays 🎄:santa: 

------------

Sync fixtures already work, with the notable caveat (HypothesisWorks/hypothesis#377) that function-scoped fixtures are only set up once for all test cases.  Automatically converting function-scoped to 'test case scoped' fixtures is possible but nontrivial.  Once that's done, supporting async function-scoped fixtures would be a few hours work plumbing it all together correctly.

Async stateful testing would be obviously hugely valuable, but there's a catch: to make it repeatable (and powerful), Hypothesis would have to control the scheduling of tasks on the event loop.  This is certainly possible, but out of scope for Hypothesis itself and I don't know enough about the asyncio internals to do it - though I'd be happy to consult on the Hypothesis side!  See e.g. HypothesisWorks/hypothesis#1693 and links; there's also a prototype `hypothesis-trio` package based on similar conversations when working on `pytest-trio` but I don't know how it's gone.

TLDR: both are difficult; not impossible but not priorities for me either.

Tinche:
Thanks for your work on this!

Issue #83, opened 2018-04-13
Clocked test case


lromor:
Hello, 
how would you achieve something similar to this:
https://asynctest.readthedocs.io/en/latest/asynctest.case.html#asynctest.ClockedTestCase

with pytest asyncio?

-l

albertodonato:
I had the same need, I'm using something like this, which is a straight copy of `ClockedTestCase` logic:

```
@pytest.fixture
async def advance_time(event_loop):

    class Clocker:

        time: float = 0

        def __init__(self, loop: asyncio.BaseEventLoop):
            self.loop = loop
            self.loop.time = lambda: self.time

        async def advance(self, seconds: float):
            if seconds < 0:
                raise ValueError('Advance time must not be negative')
            await self._drain_loop()

            target_time = self.time + seconds
            while True:
                next_time = self._next_scheduled()
                if next_time is None or next_time > target_time:
                    break

                self.time = next_time
                await self._drain_loop()

            self.time = target_time
            await self._drain_loop()

        def _next_scheduled(self):
            try:
                return self.loop._scheduled[0]._when
            except IndexError:
                return None

        async def _drain_loop(self):
            while True:
                next_time = self._next_scheduled()
                if (not self.loop._ready
                        and (next_time is None or next_time > self.time)):
                    break
                await asyncio.sleep(0)

    yield Clocker(event_loop).advance
```

in tests you can use:

```
@pytest.mark.asyncio
async def test_foo(self, advance_time):
    await do_stuff()
    await advance_time(2)
    await more_stuff()
```

albertodonato:
It would be nice to have something like this builtin

PR #110, opened 2019-02-04
Add clock event loop as default

derekbrokeit:
Extend #96 clock_event_loop fixture to become default event_loop
- uses a custom policy meta class that uses the existing policy to
  modify the new loop it creates.
- The new loop it creates is modified to provide `advance_time`
  coroutine for testing
- Extends #96 to make ClockEventLoop the default policy for testing

I am adding this as a separate merge request because I am uncertain if making this behavior default is desirable, so it is here for your consideration. As I mentioned in #96, if this behavior is to be as a separate event-loop, major refactoring to the `pytest_fixture_setup` and similar methods will be needed since they all assume `event_loop` is the name of the fixture for event-loops. There is currently no trivial way to allow custom event-loops in pytest-asyncio as far as I can tell. This is the reason #96 fails many of the tests.

### DIFF
--- a/pytest_asyncio/plugin.py
+++ b/pytest_asyncio/plugin.py
@@ -198,20 +198,23 @@ _markers_2_fixtures = {
 
 def _event_loop_policy():
     """
-    Create a new class for ClockEventLoop based on the current
-    class-type produced by `asyncio.new_event_loop()`.  This is important
+    Create a new class for ClockEventLoopPolicy based on the current
+    class-type produced by `asyncio.get_event_loop_policy()`.  This is important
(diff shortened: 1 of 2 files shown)


derekbrokeit:
Looking at this implementation, the same utilities here can actually be refactored out to a clean fixture as suggested in #83 if desired. If desired, the `advance_time` method used here could be broken out into a fixture to remove the need to change the underlying loop policy and class, which might be a cleaner solution.

derekbrokeit:
Closing this because #113 is a better solution in my opinion

Issue #95, opened 2018-09-10
[Feature Request] Event-Loop fixture that does not automatically advance time


derekbrokeit:
It would be ideal to add an additional `clock_event_loop` fixture with a custom `AbstractEventLoop`. Instead, this one would only advance time explicitly similar to the `twisted.internet.task.Clock` useful for testing `twisted` reactors. This would allow for tests to advance time and test time-based features to verify that they are correctly running when expected without having to wait for time to progress. If there is a different way to achieve this goal, I would appreciate the advice if I missed something.

albertodonato:
https://github.com/pytest-dev/pytest-asyncio/issues/83 seems similar to what you want

derekbrokeit:
Thanks @albertodonato , I've been working on this for a bit now and #110 is the most recent set of commits to achieve this behavior in case you are interested.

derekbrokeit:
Looking at your fixture in #83 is very interesting. #110 takes a different approach, changing the loop class to give all `event_loop`'s the `advance_time` coroutine, but having it as a stand alone fixture could be a good alternative solution. You would probably want to remove any calls to private loop functions though (e.g. `loop._ready`)

altendky:
Along with the inspiration from Twisted, consider Trio as well.  https://pytest-trio.readthedocs.io/en/stable/reference.html#autojump_clock  Advancing time automatically even _faster_ than the wall clock is very useful as well.

PR #113, opened 2019-02-18
add advance_time fixture and test (closes #83, #95, #96 #110)

### LINKED ISSUE #83: Clocked test case

derekbrokeit:
This is reworking of the `advance_time` method from #96 and #95 into an event-loop independent fixture. This removes the need to change the event-loop class or policy on the fly. Compared to the fixture discussed in #83, this technique does not require any private method calls to the event-loop, which potentially volatile depending on the loop implementation. One thing to keep in mind, this technique requires patching the time method of the event loop, but the original base time method is stored to reduce the assumptions required about the timing methodology, which is left up to loop creators. This should be a fairly safe implementation.

### DIFF
--- a/pytest_asyncio/plugin.py
+++ b/pytest_asyncio/plugin.py
@@ -201,6 +201,44 @@ _markers_2_fixtures = {
 }
 
 
+class EventLoopClockAdvancer:
+    """
+    A helper object that when called will advance the event loop's time. If the
+    call is awaited, the caller task will wait an iteration for the update to
+    wake up any awaiting handlers.
+    """
+    __slots__ = ("offset", "loop", "_base_time",)
+
+    def __init__(self, loop):
+        self.offset = 0.0
+        self._base_time = loop.time
+        self.loop = loop
+
+        # incorporate offset timing into the event loop
+        self.loop.time = self.time
+
+    def time(self):
+        """
+        Return the time according to the event loop's clock. The time is
+        adjusted by an offset.
+        """
+        return self._base_time() + self.offset
+
+    def __call__(self, seconds):
+        """
+        Advance time by a given offset in seconds. Returns an awaitable
+        that will complete after all tasks scheduled for after advancement
+        of time are proceeding.
+        """
+        if seconds > 0:
+            # advance the clock by the given offset
+            self.offset += abs(seconds)
+
+        # Once the clock is adjusted, new tasks may have just been
+        # scheduled for running in the next pass through the event loop
+        return self.loop.create_task(asyncio.sleep(0))
+
+
 @pytest.yield_fixture
 def event_loop(request):
     """Create an instance of the default event loop for each test case."""
@@ -237,3 +275,8 @@ def unused_tcp_port_factory():
 
         return port
     return factory
+
+
+@pytest.fixture
+def advance_time(event_loop):
+    return EventLoopClockAdvancer(event_loop)
--- a/tests/test_simple_35.py
+++ b/tests/test_simple_35.py
@@ -86,3 +86,24 @@ class Test:
 def test_async_close_loop(event_loop):
     event_loop.close()
     return 'ok'
+
+
+@pytest.mark.asyncio
+async def test_advance_time_fixture(event_loop, advance_time):
+    """
+    Test the `advance_time` fixture
+    """
+    # A task is created that will sleep some number of seconds
+    SLEEP_TIME = 10
+
+    # create the task
+    task = event_loop.create_task(asyncio.sleep(SLEEP_TIME))
+    assert not task.done()
+
+    # start the task
+    await advance_time(0)
+    assert not task.done()
+
+    # process the timeout
+    await advance_time(SLEEP_TIME)
+    assert task.done()


derekbrokeit:
I went through and added some additional tests. Also, I think the __call__ method is more clear now.

bryanforbes:
I just tried this in a Python 3.7 project I'm working on with the following code:

```python
async def race(
    futures: Iterable[FutureT[T]],
    *,
    timeout: Optional[float] = None,
    loop: Optional[asyncio.AbstractEventLoop] = None,
) -> T:
    if loop is None:
        loop = asyncio.get_running_loop()

    tasks = {future for future in futures}

    done, pending = await asyncio.wait(
        tasks, timeout=timeout, return_when=asyncio.FIRST_COMPLETED, loop=loop
    )

    try:
        if len(pending) == len(tasks):
            raise asyncio.TimeoutError()

        return done.pop().result()
    finally:
        for future in pending:
            future.cancel()
```

My test for this function looks like this:

```python
@pytest.mark.asyncio
async def test_race(event_loop, advance_time):
    async def one():
        await asyncio.sleep(100, loop=event_loop)
        return 1

    async def two():
        await asyncio.sleep(50, loop=event_loop)
        return 2

    async def three():
        await asyncio.sleep(25, loop=event_loop)
        return 3

    task = event_loop.create_task(race([one(), two(), three()], loop=event_loop))
    await advance_time(35)
    await advance_time(60)
    await advance_time(110)
    assert task.result() == 3
```

I copied the implementation in this PR into my `conftest.py`, but it didn't work until I changed the `asyncio.sleep()` calls in `EventLoopClockAdvancer.__call__` to something greater than `0`:

```python
    async def __call__(self, seconds):
        """
        Advance time by a given offset in seconds. Returns an awaitable
        that will complete after all tasks scheduled for after advancement
        of time are proceeding.
        """
        # sleep so that the loop does everything currently waiting
        await asyncio.sleep(0.0001)

        if seconds > 0:
            # advance the clock by the given offset
            self.offset += seconds

            # Once the clock is adjusted, new tasks may have just been
            # scheduled for running in the next pass through the event loop
            await asyncio.sleep(0.0001)
```

derekbrokeit:
Thanks for testing, @bryanforbes . What OS and loop policy are you loop using? I recently found that using `uvloop` causes this to fail for some reason, and I suspect the `ProactorEventLoop` may also fail. Thanks in advance for any info you can provide.

derekbrokeit:
I should mention for future reference, there is nothing (to my knowledge) we can do to fix this on `uvloop`. The problem with `uvloop` is that it does not using the public `time()` method for scheduling, it uses a private method `Loop._time()` [as seen here](https://github.com/MagicStack/uvloop/blob/master/uvloop/loop.pyx#L1275)

derekbrokeit:
@bryanforbes , you motivated me to add something I had been considering. I added a commandline option for you to set the value. Use `--advance-time-sleep 0.0001` to fix it for your use case.

bryanforbes:
> What OS and loop policy are you loop using?

I'm on macOS 10.14.1 using the standard event loop (as far as I'm aware). I updated the code in my `conftest.py` with `1e-6` and got failures in other tests. Everything worked fine with `1e-4`.

RemiCardona:
Hi folks, I'm thrilled to report that this fixture works as intended with my work codebase. Until a new release is cut out with this PR, I've vendored the fixture. This will allow us to progressively migrate our large body of tests written using asynctest (especially code using `asynctest.ClockedTestCase`), which has served us well, with neater pytest-style tests.

It's worth mentioning here (for the record) that `ClockedTestCase` works a little bit differently: time is set in stone and never "progresses". Whereas `advance_time` does not stop the time from progressing, it just allows doing controlled skips. It's too early to tell whether this subtle change will be a hindrance to our porting effort, but it is a significant difference between the two.

In any case, let me know if there's anything I can do to help making sure this PR lands. Cheers.

RemiCardona:
Hi folks, is there anything I can do to help get the ball rolling on this PR? We've been using it successfully for a few months now, it's rock solid and works as advertised. Thanks

RemiCardona:
Hi folks, I've been revisiting old TODOs at work and this issue popped up. I was thinking, to help merging this PR, maybe an "experimental" disclaimer/tag in the doc could be added to this fixture so that neither its API nor its behavior is set in stone. Could this help pave the way forward?

Maybe I should create a pytest-asyncio-advance-time package containing this fixture? (that is, until it is eventually merged, I really don't want to get into the business of maintaining yet-another tiny package) I'd really like to avoid that as pytest-asyncio really seems like the natural place for this fixture.

Any input from the maintainers would be greatly appreciated, thanks!

asvetlov:
BTW, I have a [loop proxy](https://github.com/aio-libs/aioloop-proxy) implementation that allows the wallclock time tweaking without monkey-patching of the original event loop.

Issue #112, opened 2019-02-09
Dynamically calling a fixture causes a runtime error saying "This event loop is already running"


ykuzma1:
There seems to be a bug with how [`request.getfixturevalue(argname)`](https://docs.pytest.org/en/latest/reference.html#_pytest.fixtures.FixtureRequest.getfixturevalue) interacts with pytest-asyncio. Calling the function leads to a runtime error. If you change it from being dynamically called to be fixed in the function definition, it works as expected.

Platform Info:
```
platform win32 -- Python 3.7.2, pytest-4.2.0, py-1.7.0, pluggy-0.8.1
plugins: asyncio-0.11.0.dev0
```

Test fixture that can be used in both dynamic/fixed cases:
```
@pytest.fixture
async def async_fixture():
    yield 'Hi from async_fixture()!'
```

Successful fixed function argument fixture test:
```
@pytest.mark.asyncio
async def test_async_fixture_fixed(async_fixture):
    assert async_fixture == 'Hi from async_fixture()!'
```

Failed dynamic fixture test:
```
@pytest.mark.asyncio
async def test_async_fixture_dynamic(request):
    async_fixture = request.getfixturevalue('async_fixture')
    assert async_fixture == 'Hi from async_fixture()!'
```

Failed test trace-back:
```
================================== FAILURES ===================================
_________________________ test_async_fixture_dynamic __________________________

request = <FixtureRequest for <Function test_async_fixture_dynamic>>

    @pytest.mark.asyncio
    async def test_async_fixture_dynamic(request):
>       async_fixture = request.getfixturevalue('async_fixture')

tests\test_app_factory.py:52: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
venv\lib\site-packages\_pytest\fixtures.py:478: in getfixturevalue
    return self._get_active_fixturedef(argname).cached_result[0]
venv\lib\site-packages\_pytest\fixtures.py:501: in _get_active_fixturedef
    self._compute_fixture_value(fixturedef)
venv\lib\site-packages\_pytest\fixtures.py:586: in _compute_fixture_value
    fixturedef.execute(request=subrequest)
venv\lib\site-packages\_pytest\fixtures.py:881: in execute
    return hook.pytest_fixture_setup(fixturedef=self, request=request)
[...]

PR #120, opened 2019-04-20
Fix issue #112: 'Dynamically calling a fixture causes a runtime error'

ykuzma1:
This patch works by failing over to a [`ThreadPoolExecutor`](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.ThreadPoolExecutor) in cases where the expected event loop is already running. Notably this happens when calling [`request.getfixturevalue(argname)`](https://docs.pytest.org/en/latest/reference.html#_pytest.fixtures.FixtureRequest.getfixturevalue) because it dynamically calls a fixture that needs to be setup on an event loop that's already running the asynchronous `pytest` tests.

### DIFF
--- a/pytest_asyncio/plugin.py
+++ b/pytest_asyncio/plugin.py
@@ -1,5 +1,6 @@
 """pytest-asyncio implementation."""
 import asyncio
+import concurrent.futures
 import contextlib
 import functools
 import inspect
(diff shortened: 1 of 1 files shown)


ykuzma1:
Sorry, I was a bit premature. I was hoping to avoid changing the behavior of the existing code as much as possible by `try`ing to run `setup()` as before and only falling back to running within the `Executor` if a `RuntimeException` occurred. Unfortunately, this was running `setup()` multiple times (🤦‍♂️) causing issues with test setup and tear-down. This means that it has to run `setup()` in the `Executor` no matter what.

The only other option is detecting whether the fixture was called dynamically from within `pytest_fixture_setup()`, but I'm not aware of any way to do that.

simonfagerholm:
@ykuzma1 How about adding a marker that can be identified from within the hooks?
So if you know that your testcase needs to call fixture dynamically you can mark that testcase

asvetlov:
I guess the PR should be closed.
Running an async fixture setup phase in a thread pool violates the asyncio usage rules.
Thanks for the PR anyway!

Issue #117, opened 2019-03-27
Future <Future pending> attached to a different loop error if hypothesis and @pytest.mark.asyncio are used


VladimirWork:
Hello, could you please help me with this issue? I'm trying to send values from hypothesis in async test with await calls inside and getting subject error (I've tried trick from #38 for event_loop but it still doesn't work, also it looks like your hypothesis tests have no awaits inside but asserts only so could you point me to any real example of hypothesis+pytest-asyncio usage?):

platform linux -- Python 3.5.2, pytest-4.3.1, py-1.8.0, pluggy-0.7.1
plugins: hypothesis-4.14.0, testinfra-2.0.0, asyncio-0.10.0

conftest.py:

> @pytest.fixture(scope="session")
> def event_loop():
>     loop = asyncio.get_event_loop()
>     loop.run_until_complete(pool.set_protocol_version(2))
>     yield loop
>     loop.close()

test.py:

> @settings(verbosity=Verbosity.debug)
> @given(test_string=st.characters(whitelist_categories=('N', 'L')))
> @pytest.mark.asyncio
> async def test_misc_hypothesis(event_loop, test_string):
>     pool_handle, _ = await pool_helper()
>     wallet_handle, _, _ = await wallet_helper()
>     trustee_did, _ = await default_trustee(wallet_handle)
>     res = await schema_helper(pool_handle, wallet_handle, trustee_did, test_string, '1.0',
>                               json.dumps(["age", "sex", "height", "name"]))
>     print(res)

stacktrace:

> =================================== FAILURES ===================================
> _____________________________ test_misc_hypothesis _____________________________
> 
> event_loop = <_UnixSelectorEventLoop running=False closed=False debug=False>
> 
>     @settings(verbosity=Verbosity.debug)
> >   @given(test_string=st.characters(whitelist_categories=('N', 'L')))
>     @pytest.mark.asyncio
>     async def test_misc_hypothesis(event_loop, test_string):
> 
> test_misc.py:623: 
> _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
> /usr/lib/python3.5/contextlib.py:77: in __exit__
>     self.gen.throw(type, value, traceback)
> /usr/local/lib/python3.5/dist-packages/hypothesis/_settings.py:402: in local_settings
>     yield s
[...]

PR #122, opened 2019-05-04
Sync wrapper for hypothesis tests uses the event loop provided by pytest-asyncio

seifertm:
Async test functions marked for execution by Hypothesis are currently wrapped in a synchronous function during the test discovery phase of pytest. At this point, the `event_loop` fixture provided by pytest-asyncio has not been evaluated and the wrapper is forced to create a new loop for each hypothesis test. This may result is problems when using async-related code in fixtures as in #117.

This PR moves the wrapping code to the actual test execution. At this point the `event_loop` fixture is evaluated and there is no need to create a new loop for the test.

What's missing from this PR is a proper automated test case. Since I have run into this issue using aiohttp, I was using the following code to test it:
```python
import aiohttp
import pytest
from hypothesis import given
from hypothesis.strategies import integers


@pytest.fixture
async def http():
    async with aiohttp.ClientSession() as session:
        yield session


@pytest.mark.asyncio
@given(integers())
async def test_connect(http, n):
    await http.get('http://127.0.0.1:8080', timeout=None)
```
The test succeeds with the patch and fails otherwise.

Currently, one of the test cases still fails, so this cannot be merged. I would like to hear your comments on this, though! Especially from @Zac-HD as the original author of the hypothesis integration and @VladimirWork to see if this patch also solves his problem described in #117.

### DIFF
--- a/pytest_asyncio/plugin.py
+++ b/pytest_asyncio/plugin.py
@@ -51,28 +51,35 @@ def pytest_pycollect_makeitem(collector, name, obj):
 @pytest.hookimpl(hookwrapper=True)
 def pytest_fixture_setup(fixturedef, request):
     """Adjust the event loop policy when an event loop is produced."""
+    if fixturedef.argname == "event_loop" and 'asyncio' in request.keywords:
+        outcome = yield
+        loop = outcome.get_result()
+        policy = asyncio.get_event_loop_policy()
+        try:
+            old_loop = policy.get_event_loop()
+        except RuntimeError as exc:
+            if 'no current event loop' not in str(exc):
+                raise
+            old_loop = None
+        policy.set_event_loop(loop)
+        fixturedef.addfinalizer(lambda: policy.set_event_loop(old_loop))
+        return
+
     if isasyncgenfunction(fixturedef.func):
         # This is an async generator function. Wrap it accordingly.
-        f = fixturedef.func
+        generator = fixturedef.func
 
-        strip_event_loop = False
-        if 'event_loop' not in fixturedef.argnames:
-            fixturedef.argnames += ('event_loop', )
-            strip_event_loop = True
         strip_request = False
         if 'request' not in fixturedef.argnames:
             fixturedef.argnames += ('request', )
             strip_request = True
 
         def wrapper(*args, **kwargs):
-            loop = kwargs['event_loop']
             request = kwargs['request']
-            if strip_event_loop:
-                del kwargs['event_loop']
             if strip_request:
                 del kwargs['request']
 
-            gen_obj = f(*args, **kwargs)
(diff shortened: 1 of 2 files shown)


pytest_asyncio/plugin.py
```diff
@@ -145,10 +145,17 @@ def pytest_pyfunc_call(pyfuncitem):
...
-
+            if getattr(pyfuncitem.obj, 'is_hypothesis_test', False):
+                # Wrap the async function to execute in the pytest-asyncio event loop and yield to delegate
+                # test execution to the hypothesis plugin
+                pyfuncitem.obj.hypothesis.inner_test = wrap_in_sync(
+                    pyfuncitem.obj.hypothesis.inner_test, event_loop
+                )
+                # It is important to yield, not return. Otherwise the event_loop fixture will get cleaned up too early
+                # when using a combination of hypothesis and pytest.mark.parametrize
+                yield None
```
seifertm: As the comment says, `return None` will cause problems with parametrized hypothesis tests. However, changing `return None` to `yield None` breaks `test_simple::test_asyncio_marker_fail` for no apparent reason. Do you have any idea why this might cause `xfail` to break?

Zac-HD:
I can't put my finger on the exact problem, but I have a bad feeling about isolation and the timing of cleanup.  That is, that the event_loop might live too long *and* there might be cases where it gets closed between runs of the test case.  The yield-not-return thing makes me super suspicious for this reason.

seifertm:
You are right: The "yield-not-return" thing is smelly and affects other pytest markers, so I got rid of it. At this point, the parametrization is broken, because the event loop is already closed after the first parameter.

This is weird, because `pytest -vv --setup-show -k mark_and_parametrize tests/` shows that the `event_loop` is correctly set up and torn down before and after each test item. I spent quite some time debuggin, but can't seem to find the issue here…

Thanks to @TimothyFitz who added a test case that reproduces the issue!

seifertm:
All tests run successfully now and the issue with the prematurely closed event loop seems to be fixed.

I did this by simplifying the wrapper for async functions: `pytest_runtest_setup` ensures that all tests marked with `asyncio` require the `event_loop` fixture and `pytest_fixture_setup` replaces the current event loop with the loop provided by the fixture. This means, if we call `get_event_loop()`, we can assume that we get the loop provided by the pytest-asyncio fixture. I did exactly that in the async wrapper.

seifertm:
I also added two refactoring commits that hopefully make the code easier to understand.

Zac-HD:
Congratulations!  This looks great to me, and the refactoring makes it much easier to follow.

Zac-HD:
Congratulations!  This looks great to me, and the refactoring makes it much easier to follow.

Tinche:
LGTM!

PR #128, opened 2019-10-02
Fixing file encoding per PEP-263

thirtytwobits:
Python source defaults to ASCII without encoding hints.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,3 +1,4 @@
+# -*- coding: utf-8 -*-
 import re
 from pathlib import Path
 


eirnym:
Default encoding for python 3+ is `utf-8`. Why do you need to add an encoding specifier for that?

thirtytwobits:
Huh, that's a great question. I added this because an automated system I encountered when using pytest-asyncio wouldn't let me proceed without it. I then failed to read the applicability of PEP-263 and assumed this was a bug in your library. I suppose the bug is with that tool. Sorry. RTFM me :)

PR #140, opened 2020-02-12
Update plugin.py to enable asyncio debugging

nekokatt:
Changed plugin.py to enable debug on the event loop fixture.

This allows any tasks that are not closed correctly to be logged 
in such a way that we can see which test they originate from, 
which is incredibly useful in testing large libraries where a test
somewhere may not be awaiting something correctly or may be
closing the event loop without terminating tasks correctly.

### DIFF
--- a/pytest_asyncio/plugin.py
+++ b/pytest_asyncio/plugin.py
@@ -170,6 +170,7 @@ def pytest_runtest_setup(item):
 def event_loop(request):
     """Create an instance of the default event loop for each test case."""
     loop = asyncio.get_event_loop_policy().new_event_loop()
+    loop.set_debug(True)
     yield loop
     loop.close()
 


simonfagerholm:
@nekokatt Maybe it should be possible to turn this off and on, debugging code permanently on could possibly annoy users.

nekokatt:
Arent these warnings almost always an indicator that your tests have issues that can affect how they operate, though?

simonfagerholm:
That is possible, I was reflecting over that it is an option on the loop and not the default behavior.
If it raises new warnings for people that don't care about them I think they will start creating issues about it.

Will this not also show warnings from other parts except the tests, like all asyncio code?

Tinche:
How does this relate to setting the environment variable PYTHONASYNCIODEBUG = 1?

simonfagerholm:
Good point, it seems this can be achieved without change in pytest-asyncio by setting the environment variable in system where it is wanted

asvetlov:
The PR should be closed I guess.

Issue #141, opened 2020-03-12
pytest 5.4.0 deprecated `pytest.Function()` construction


mjpieters:
See the [5.4.0 list of deprecations](https://docs.pytest.org/en/5.4.0/changelog.html#deprecations):

> #5975: Deprecate using direct constructors for Nodes.
>
> Instead they are new constructed via `Node.from_parent`.

and pytest-dev/pytest#5975.

This results in a failure to collect tests, see [this example CI run](https://dev.azure.com/aio-libs/aiosignal/_build/results?buildId=2511&view=logs&j=b49f96fc-8ec4-5547-7cfa-13f1a7414ddb&t=1f0169f8-ee0c-557e-d55c-713877620c33):

```python
____________________ ERROR collecting tests/test_signals.py ____________________
/opt/hostedtoolcache/Python/3.8.2/x64/lib/python3.8/site-packages/pluggy/hooks.py:286: in __call__
    return self._hookexec(self, self.get_hookimpls(), kwargs)
/opt/hostedtoolcache/Python/3.8.2/x64/lib/python3.8/site-packages/pluggy/manager.py:93: in _hookexec
    return self._inner_hookexec(hook, methods, kwargs)
/opt/hostedtoolcache/Python/3.8.2/x64/lib/python3.8/site-packages/pluggy/manager.py:84: in <lambda>
    self._inner_hookexec = lambda hook, methods, kwargs: hook.multicall(
/opt/hostedtoolcache/Python/3.8.2/x64/lib/python3.8/site-packages/_pytest/python.py:216: in pytest_pycollect_makeitem
    res = outcome.get_result()
/opt/hostedtoolcache/Python/3.8.2/x64/lib/python3.8/site-packages/pytest_asyncio/plugin.py:39: in pytest_pycollect_makeitem
    item = pytest.Function(name, parent=collector)
/opt/hostedtoolcache/Python/3.8.2/x64/lib/python3.8/site-packages/_pytest/nodes.py:83: in __call__
    warnings.warn(NODE_USE_FROM_PARENT.format(name=self.__name__), stacklevel=2)
E   pytest.PytestDeprecationWarning: direct construction of Function has been deprecated, please use Function.from_parent
```

PR #142, opened 2020-03-13
plugin: Use pytest 5.4.0 new Function API

romainletendart:
This fixes https://github.com/pytest-dev/pytest-asyncio/issues/141.

### DIFF
--- a/pytest_asyncio/plugin.py
+++ b/pytest_asyncio/plugin.py
@@ -36,13 +36,13 @@ def pytest_configure(config):
 def pytest_pycollect_makeitem(collector, name, obj):
     """A pytest hook to collect asyncio coroutines."""
     if collector.funcnamefilter(name) and _is_coroutine(obj):
-        item = pytest.Function(name, parent=collector)
+        item = pytest.Function.from_parent(collector, name=name)
 
         # Due to how pytest test collection works, module-level pytestmarks
         # are applied after the collection step. Since this is the collection
         # step, we look ourselves.
         transfer_markers(obj, item.cls, item.module)
-        item = pytest.Function(name, parent=collector)  # To reload keywords.
+        item = pytest.Function.from_parent(collector, name=name)  # To reload keywords.
 
         if 'asyncio' in item.keywords:
             return list(collector._genfunctions(name, obj))
--- a/setup.py
+++ b/setup.py
@@ -40,7 +40,7 @@ setup(
         "Framework :: Pytest",
     ],
     python_requires=">= 3.5",
-    install_requires=["pytest >= 3.6.0, < 5.4.0"],
+    install_requires=["pytest >= 5.4.0"],
     extras_require={
         ':python_version == "3.5"': "async_generator >= 1.3",
         "testing": [


mjpieters:
This change means this project can only work with Pytest 5.4.0 and up. Either

- Update the install_requires metadata to reflect this dependency
- Allow for a fallback, perhaps with a `try: ... except AttributeError:` block, to calling `pytest.Function()` directly again.

romainletendart:
@mjpieters Thank you for your review.
I have updated `install_requires` accordingly.

simonfagerholm:
@romainletendart: The reason this PR fails has been fixed in #148, just waiting for review (maybe @mjpieters has the time to check?) and merge.
This branch will probably need to pull in the master branch when that happens, as the changes are needed and that my commit also changed the install_requires, but to only work with pytest < 5.4.0.

dmitrypolo:
@simonfagerholm wouldn't it be better to follow @mjpieters advice and instead of constraining users from using newer versions of `pytest` to have a fallback to call `pytest.Function()` based on the version?

simonfagerholm:
@dmitrypolo maybe you meant  to message @romainletendart?

dmitrypolo:
@simonfagerholm  @romainletendart my comment applies to both of you technically. This PR confines users from using `pytest < 5.4` whereas yours confines users from using `pytest > 5.4`.

simonfagerholm:
@dmitrypolo Before this PR is merged pytest 5.4.0 and above is not supported. To be able to use pytest >= 5.4.0 this PR is needed. My PR correctly reflects that only pytest < 5.4.0 is supported

dmitrypolo:
@simonfagerholm correct, I guess the point I am trying to get across is we should probably try to allow for both users of `pytest < 5.4` and `pytest > 5.4` to use this plug-in. The issue is that both of these PRs restrict users in one-way or the other.

simonfagerholm:
@dmitrypolo Yes but in my case its for good reason: It isn't compatible... I can understand that you would like it to be backwards compatible, but that is only with this PR and not mine.

romainletendart:
> @simonfagerholm wouldn't it be better to follow @mjpieters advice and instead of constraining users from using newer versions of `pytest` to have a fallback to call `pytest.Function()` based on the version?

I'm not sure supporting multiple API versions inside the code itself is a good idea.
Just suppose you do this for every API change and your code quickly becomes unreadable because of all the ifs you added to support multiple versions.
Plus, you get high chances to get corner case bugs with some weird combinations of versions that were not supposed to work together initially.

@simonfagerholm, I will definitely rebase on the upstream master once your merged your PR just to be sure everything is green before being able to merge my own PR.

simonfagerholm:
@romainletendart I totally agree.
I'll let you know when its merged. Just need someone with write access to review and approve.

romainletendart:
Maybe we could ask @Tinche as he seems to be both the initiator and the main contributor of pytest-asyncio?

Tinche:
@romainletendart Fix the merge conflict please and let's get this in!

simonfagerholm:
@romainletendart Like @Tinche said: my PR is in, pull it and lets get this in! :)

romainletendart:
@Tinche @simonfagerholm Merge conflict fixed, thank you! :)

simonfagerholm:
@Tinche Seems good to go! :)

Tinche:
Thank you kindly!

PR #138, opened 2020-01-23
Replace yield_fixture() by fixture()

atugushev:
`@pytest.yield_fixture` is deprecated since pytest-3.0, see:
https://docs.pytest.org/en/latest/yieldfixture.html

PR #146, opened 2020-03-21
don't use deprecated pytest.yield_fixture in readme

Mattwmaster58:
See https://docs.pytest.org/en/latest/yieldfixture.html:

> ## Important
>
> Since pytest-3.0, fixtures using the normal fixture decorator can use a yield statement to provide fixture values and execute teardown code, exactly like yield_fixture in previous versions.
>
> Marking functions as yield_fixture is still supported, but deprecated and should not be used in new code.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -81,7 +81,7 @@ event loop. This will take effect even if you're using the
 
 .. code-block:: python
 
-    @pytest.yield_fixture()
+    @pytest.fixture()
     def event_loop():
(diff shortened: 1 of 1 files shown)


simonfagerholm:
Depends on #142 and #148 being merged to pass builds

simonfagerholm:
Duplicate of #138

Tinche:
Merged in the other one, thanks!

PR #148, opened 2020-03-23
Fix unittests and add max supported pytest version

### LINKED ISSUE #145: Hypothesis test fail because of change in Hypothesis
See PR #142 where it first occurs, but it will happen on all builds after last of feb because of https://github.com/HypothesisWorks/hypothesis/pull/2356.

@Zac-HD maybe you can help with a fix as you are involved in both this and hypothesis package.

Basically the problem is that the even_loop is function scope.

simonfagerholm:
This PR contains 3 changes:
* Fix for running test_subprocess on windows by following instructions on https://docs.python.org/3.7/library/asyncio-subprocess.html#creating-subprocesses
* Change event_loop-fixture to module-scope in hypothesis tests, fixes #145 
* Change setup.py max version of pytest to avoid failures because of changes in 5.4.0 until #141 merged. It should be reverted in that PR.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -46,7 +46,7 @@ setup(
         "testing": [
             "coverage",
             "async_generator >= 1.3",
-            "hypothesis >= 3.64",
+            "hypothesis >= 5.7.1",
         ],
     },
     entry_points={"pytest11": ["asyncio = pytest_asyncio.plugin"]},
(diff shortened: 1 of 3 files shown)


simonfagerholm:
@Tinche: Could you have a look at this? It's fixing a problem causing build failures in other PRs

Tinche:
Thanks!

simonfagerholm:
Thanks for he merge, @Tinche!

Issue #127, opened 2019-08-09
ContextVar not propagated from fixture to test


beedub:
I'm trying to figure out why this code sample isn't working. I'm not experienced with asyncio yet, so this is very likely a simple misunderstanding on my part.

```python
from _contextvars import ContextVar

import pytest

blah = ContextVar("blah")


@pytest.fixture
async def my_context_var():
    blah.set("hello")
    assert blah.get() == "hello"
    yield blah


@pytest.mark.asyncio
async def test_blah(my_context_var):
    assert my_context_var.get() == "hello" # this fails
```

eirnym:
Short answer: You probably want to do a `return` instead of `yield` on last line of `my_context_var`.

Explanation:
If an async function contains `yield`, it  becomes a generator as well. Thus you should threat it as generator. The return value of generator exists since Python 3.3, so you can check it as well if it's required. For more information on returning values, please read [PEP 380](https://www.python.org/dev/peps/pep-0380) on how return values from generator and [PEP 479](https://www.python.org/dev/peps/pep-0479) to not use `StopIteration`

kalefranz:
I've also hit this. I think what's happening is explained by

https://github.com/pytest-dev/pytest-asyncio/blob/2c8023d5e123644a1990dcf978b527ddafeca8f7/pytest_asyncio/plugin.py#L97

along with understanding of https://www.python.org/dev/peps/pep-0550.  The `setup()` function there executes the code within the coroutine fixture.  Using `loop.run_until_complete(setup())` means that context changes won't "leak" outside of the execution context within the coroutine.

One option is to try to avoid using a coroutine fixture when you need to set context.  So for example

```
import asyncio
from contextvars import ContextVar
import pytest

blah = ContextVar("blah")


async def set_context():
    blah.set("hello")
    assert blah.get() == "hello"


@pytest.fixture
def my_context_var():
    asyncio.run(set_context())
    assert blah.get() == "hello"
    yield blah


@pytest.mark.asyncio
async def test_blah(my_context_var):
    assert my_context_var.get() == "hello"
```

Andrei-Pozolotin:
+1

nolar:
A note on the workaround above: the `set_context()` coroutine will run on its own, and its contextvar will not get into the surrounding/subling coroutines. The example above fails with:

```
__________________________________________________________________________ ERROR at setup of test_blah ___________________________________________________________________________

    @pytest.fixture
    def my_context_var():
[...]

PR #153, opened 2020-04-18
add support for async contextvars

n1ywb:
this ugly hack causes the context to be preserved across all excursions into async land. this fixes https://github.com/pytest-dev/pytest-asyncio/issues/127


n1ywb:
works on python 3.8 at least 😅

simonfagerholm:
Why are the hypothesis tests removed?

n1ywb:
> Why are the hypothesis tests removed?

oops

n1ywb:
> Why are the hypothesis tests removed?

honestly its because my internet went down last night and I couldn't install the hypothesis package at the time

n1ywb:
one caveat I might mention about this implementation is that EVERY async task will run in the same context. normally child tasks would inherit a COPY of the parent task's context. but it's probably an acceptable trade off in a test environment. but it could cause context to unexpectedly leak from a child task into a parent task.

n1ywb:
I opened https://bugs.python.org/issue40320 to officially request the ability to pass context to task

achimnol:
I just hit [a similar problem regarding contextvars with async-yield fixtures](https://aio-libs.discourse.group/t/contextvars-with-pytest-async-generator-fixture/92).
Could your PR also resolve this case as well?

n1ywb:
> I just hit [a similar problem regarding contextvars with async-yield fixtures](https://aio-libs.discourse.group/t/contextvars-with-pytest-async-generator-fixture/92).
> Could your PR also resolve this case as well?

It should. I use them in my app tests which is the reason I did this.

jpic:
Works great, just had to rebase it because master changed from doing "append if not in item.fixturenames" with "remove if in item.fixturenames then insert".

Thank you very much! :tophat:

lephuongbg:
Verified that this one works. I like this one more than #161 since I can easily copy the necessary parts to my `conftest.py`.

mr-rodgers:
Is this fix abandoned?

seifertm:
I really hate to do this, but I'm closing this PR in favour of the solution proposed in #312. I'll my best to outline my reasons.

I like that you found a nice _Task_ abstraction. It has a high "code locality" (i.e. the functionality coherent and found mostly inside `Task`). The PR also provides the appropriates tests.

However, the `Task` class makes use of protected types, such as `asyncio.tasks._PyTask` and `asyncio.futures._PyFuture`. Their names or functionality might change between patch releases. As we only test with different minor versions of Python, we likely wouldn't be able to discover any breakage.

The PR also introduces the `context` fixture, which I see as problematic. Pytest-asyncio currently encourages overriding the `event_loop` fixture, in order to control the fixture scope. Unfortunately, people tend to write all sorts of code into their `event_loop` fixtures, in order to "inject" functionality into their tests. This makes it hard to maintain pytest-asyncio as the custom code in the loop fixture causes additional corner cases that break when changing the plugin code. Therefore, we should try to reduce the number of fixtures provided by pytest-asyncio, rather than adding more (see #311). I fear  that adding the `context` fixture will open another can of worms similar to `event_loop`.

Python 3.11 introduces [asyncio.Runner](https://docs.python.org/3.11/library/asyncio-runner.html#asyncio.Runner), which essentially allows running multiple async functions in the same context. #312 introduces the concept of a _Runner_ in pytest-asyncio. I think this is the approach we should take

Others have reported they could use similar code in their conftest.py, so this may serve them as a workaround, until #127 is fixed.

I'm sorry this PR was stale for such a long time. A single maintainer can only spend so much time and we only "added staff" in the beginning of this year.

PR #172, opened 2020-06-28
Fix #168: 1) respect user case that provides specific event loop policy, 2) add a new unitary test to check it

### LINKED ISSUE #168: Release 0.14 breaks tests that rely on a specific event_loop_policy
# Problem

We have many testcases that rely on `asyncio.set_event_loop_policy(AnyThreadEventLoopPolicy())`
We set the event loop policy in the conftest.py, such that it is set before any tests start.

Since 0.14 these testcases produce errors 

```
RuntimeError: There is no current event loop in thread 'ThreadPoolExecutor-109_6'.
```

# Root cause 

The change here: https://github.com/pytest-dev/pytest-asyncio/pull/164/files#diff-4d85e8313d8e67cb777137411a11ac86R80

Resets the event loop policy after the first test, breaking all the next tests.

alblasco:
Hi @Tinche 
Let me know if this PR can be helpful to support the project.
Please let me know your thoughts or share the solution you are working on.

Don't hesitate to let me know what's the best way to help

### DIFF
--- a/pytest_asyncio/plugin.py
+++ b/pytest_asyncio/plugin.py
@@ -72,12 +72,6 @@ class FixtureStripper:
             del data_dict[name]
         return result
 
-@pytest.hookimpl(trylast=True)
-def pytest_fixture_post_finalizer(fixturedef, request):
-    """Called after fixture teardown"""
-    if fixturedef.argname == "event_loop":
-        # Set empty loop policy, so that subsequent get_event_loop() provides a new loop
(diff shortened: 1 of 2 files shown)


alblasco:
Hi @Tinche 
After submitting PR, I have just realised you made a policies branch. 

Apologise. Let me know your feedback

Tinche:
Let's move the discussion to https://github.com/pytest-dev/pytest-asyncio/pull/174

PR #173, opened 2020-06-28
Fix new tests

alblasco:
The only problem with new tests are that event_loop fixture is executed before event_loop_policy fixture. 

Adding a first test (not async), depending on event_loop_policy fixture, forces to execute it the first.


Tinche:
Yeah, this will fix the tests but doesn't actually fix the issue, so we would have to document this. I'm going to open a pull request for the `policies` branch, let's move the discussion there.

Issue #154, opened 2020-04-21
Error of "attached to a different loop" appears in 0.11.0 but not in 0.10.0


krizex:
My testcase get the error of "attached to a different loop" in 0.11.0 but works fine in 0.10.0

The detail is I have the following code in my testcase:
```
asyncio.get_event_loop().run_in_executor(...)
```

Is there any related change in 0.11.0?

A simple testcase for reproducing:
```
import asyncio
import pytest

@pytest.fixture(scope='function')
async def loop():
    return asyncio.get_event_loop()

def foo():
    return 0

@pytest.mark.asyncio
async def test_async(loop):
    await loop.run_in_executor(None, foo)
```

simonfagerholm:
@krizex Can you provide the shortest possible example that reproduces this?

krizex:
@simonfagerholm Thanks for you quick response! Updated the testcase in my original post, please check.

elemoine:
We're getting hit by the same issue. We need to stick to 0.10.0 for now.

Gr1N:
Got the same issue with 0.11.0...

simonfagerholm:
I tested a little bit, it seems the `event_loop `fixture is not started before the fixture `foo `in your example. That means that when `event_loop` is started it will create a new loop to run the test.
The behaviour seems to have changed a bit in the plugin hooks, I suspect `pytest_fixture_setup` might be the culprit.

By including `event_loop `as a dependency of `foo` the example test works.

Also the loop provided by `event_loop` can be used directly like this:
```
import pytest

def foo():
    return 0

@pytest.mark.asyncio
async def test_async(event_loop):
    await event_loop.run_in_executor(None, foo)
```

Would that work as a work around?

krizex:
The example testcase is just for reproducing the issue. My real usecase is not that simple and use `asyncio.get_event_loop` in some low level module so I could not leverage the `event_loop` fixture.

At this moment I will stick to v0.10.0 until the issue been fixed.

simonfagerholm:
Is it possible to make your fixture dependant on the `event_loop`, i.e add `event_loop ` as an argument to the fixture?

Gr1N:
It will be not only about tests to rewrite. Because of that you somehow need to pass `event_loop` fixture in your production code...

simonfagerholm:
No, as long as the `event_loop` fixture is started before the production code is executed `get_event_loop` should return the loop from `event_loop`

[...]

PR #174, opened 2020-06-28
Alternative policies support

Tinche:
Introduces code for handling custom event loop policies and tests for it.

Apologies for reformatting noise, I'm using black now.

### DIFF
--- a/pytest_asyncio/plugin.py
+++ b/pytest_asyncio/plugin.py
@@ -6,12 +6,15 @@ import inspect
 import socket
 
 import pytest
+
 try:
     from _pytest.python import transfer_markers
 except ImportError:  # Pytest 4.1.0 removes the transfer_marker api (#104)
+
     def transfer_markers(*args, **kwargs):  # noqa
         """Noop when over pytest 4.1.0"""
         pass
+
 
 try:
     from async_generator import isasyncgenfunction
@@ -26,10 +29,12 @@ def _is_coroutine(obj):
 
 def pytest_configure(config):
     """Inject documentation."""
-    config.addinivalue_line("markers",
-                            "asyncio: "
-                            "mark the test as a coroutine, it will be "
-                            "run using an asyncio event loop")
+    config.addinivalue_line(
+        "markers",
+        "asyncio: "
+        "mark the test as a coroutine, it will be "
+        "run using an asyncio event loop",
+    )
 
 
 @pytest.mark.tryfirst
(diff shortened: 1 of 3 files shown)


Tinche:
The tests are failing atm because in the test scenario and in the toy project I've created to test this the event_loop fixture gets executed before the `event_loop_policy` fixture for the first test only. This could be because we inject the event_loop fixture into the list of fixtures as the first element. So maybe the logic there could be improved.

Tinche:
One fix is to override the `event_loop` fixture to depend on your `event_loop_policy` fixture explicitly:

```
@pytest.fixture(autouse=True, scope="session")
def event_loop_policy():
    asyncio.set_event_loop_policy(CustomSelectorLoopPolicy())
    yield
    asyncio.set_event_loop_policy(None)


@pytest.fixture()
def event_loop(event_loop_policy):
    return asyncio.new_event_loop()
```
This is boilerplate though, let's see if we can come up with something better.

Tinche:
Fun fact: the `event_loop` fixture being prepended instead of appended to the list of fixtures for asyncio tests was added recently in https://github.com/pytest-dev/pytest-asyncio/issues/154.

Tinche:
The solution I'm leaning towards is making `event_loop_policy` a magical fixture you should override if you want to install your own policy, like `event_loop` is a magical fixture.

alblasco:
> Apologies for reformatting noise, I'm using black now.

I use black too (and prospector)


> The solution I'm leaning towards is making `event_loop_policy` a magical fixture you should override if you want to install your own policy, like `event_loop` is a magical fixture.

This sounds good. What about the following, as an easy solution :
* Option 1: Only one fixture. You have to include inside event_loop:
```
@pytest.fixture
def event_loop():
    asyncio.set_event_loop_policy(XXXLoopPolicy())
    yield loop
    loop.close()
    asyncio.set_event_loop_policy(None)
```
* Option 2: Two fixtures (in very special cases, if any, as normally it wouldn´t be needed). And then users have to make explicit dependencies:
```
@pytest.fixture
def event_loop_policy():
    asyncio.set_event_loop_policy(XXXLoopPolicy())
    yield loop
    asyncio.set_event_loop_policy(None)

@pytest.fixture
def event_loop(event_loop_policy):
    yield loop
    loop.close()
```
Or do you want to request a mandatory event_loop_policy fixture?. This could be more disruptive.

Tinche:
Hm actually the suggestion of having users override the policy in the loop fixture is simpler, but less obvious. That might be the way to go if we document it properly. Ideally you wouldn't need to override the policy, overriding the loop would be enough (although you might need to override the policy if the code you're testing deals with it directly).

alblasco:
So I understand that option 2 is the way to go. Fine, it sounds good to me.
Let me know if I can help in anyway.

Tinche:
@alblasco Actually I was going for 1.a. Seems the simplest, we just need to document it.

alblasco:
> @alblasco Actually I was going for 1.a. Seems the simplest, we just need to document it.

Python ZEN: Simple is better than complex.

seifertm:
@Tinche `pytest-asyncio` no longer resets the current policy when creating a new event loop. That means users could set loop policies using regular pytest fixtures.

Can you think of another use case covered by this PR? Is this something we want to pursue?

asvetlov:
Agree with @seifertm

Tinche:
@seifertm agreed! thanks

PR #190, opened 2020-10-05
Add official Python 3.9 support

seifertm:
Tests run successfully on Python 3.9 without modifications. Added it to the list of supported Python interpreters in `setup.py` and to the list of tested environments in tox and TravisCI.


Tinche:
Thanks!

PR #192, opened 2020-10-05
Teardown of the event_loop fixture no longer replaces the event loop policy

### LINKED ISSUE #168: Release 0.14 breaks tests that rely on a specific event_loop_policy

### LINKED ISSUE #188: 0.14 overrides event loop choice and breaks tests
Bokeh relies on Tornado which is incompatible with the default Proactor event loop on Windows starting with Python 3.8. 

Accordingly, we call this function
```
def fixup_windows_event_loop_policy() -> None:
    if sys.platform == 'win32' and sys.version_info[:3] >= (3, 8, 0):
        import asyncio
        if type(asyncio.get_event_loop_policy()) is asyncio.WindowsProactorEventLoopPolicy:
            # WindowsProactorEventLoopPolicy is not compatible with tornado 6
            # fallback to the pre-3.8 default of WindowsSelectorEventLoopPolicy
            asyncio.set_event_loop_policy(asyncio.WindowsSelectorEventLoopPolicy())

```
in any module that will need to use tornado. Since 0.14 became available, all our Tornado tests are failing, because the Proactor loop is being used in tests even despite our explicit event loop policy selection:
```

509
self = <ProactorEventLoop running=False closed=False debug=False>, fd = 2832
510
callback = <bound method BaseAsyncIOLoop._handle_events of <tornado.platform.asyncio.AsyncIOMainLoop object at 0x0000020E9CBDCB50>>
511
args = (2832, 1)
```

Is this intentional? is there some new way we should configure pytest-asyncio to be able to not override our policy selection?

ref: https://github.com/bokeh/bokeh/issues/10514

seifertm:
My understanding is that everyone is happy, if the event loop policy remains unaltered. I think I found a way to get rid of the code that resets the event loop policy.

## Problem
Here is my understanding of the problem:
We need to make sure that each test case runs in an isolated fashion. Therefore, the `event_loop` fixture provides a fresh event loop and is automatically applied to all marked tests. When the fixture goes out of scope, it closes the event loop. Subsequent test cases would run into the following error:
```
RuntimeError: Event loop is closed
```

This is prevented a dedicated [fixture teardown](https://github.com/pytest-dev/pytest-asyncio/blob/4e2e6bd16682f6d3f8597cf5d7a86521a847e7ba/pytest_asyncio/plugin.py#L75-L80) prodecure. The teardown sets the event loop policy to `None`, which also disposes the existing, closed event loop. Subsequent calls to `asyncio.get_event_loop()` will return a fresh loop. The problem is that setting the event loop policy to None resets the policy to the system default, which may be undesired.

## Solution
Instead of resetting the event loop policy, the same effect can be achieved by requesting a new loop to replace the existing, closed event loop:
```
@pytest.hookimpl(trylast=True)
def pytest_fixture_post_finalizer(fixturedef, request):
    """Called after fixture teardown"""
    if fixturedef.argname == "event_loop":
        new_loop = policy.new_event_loop()  # Replace existing event loop
        policy.set_event_loop(new_loop)  # Ensure subsequent calls to get_event_loop() succeed
```

## Considerations
What I dislike about this solution is that a new loop is created during the _teardown_ of the `event_loop` fixture. It feels wrong to do it here. The correct place would be to create a new loop during fixture setup. In fact, the fixture _does_ create another loop so we effectively create two new event loops for each test run. This points to an issue with the order in which the fixtures are evaluated.

What are your thoughts on this approach?

Resolves #168
Resolves #188

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -164,6 +164,10 @@ Only test coroutines will be affected (by default, coroutines prefixed by
 
 Changelog
 ---------
+0.17.0 (UNRELEASED)
+~~~~~~~~~~~~~~~~~~~
+- `pytest-asyncio` no longer alters existing event loop policies. `#168 <https://github.com/pytest-dev/pytest-asyncio/issues/168>`_, `#188 <https://github.com/pytest-dev/pytest-asyncio/issues/168>`_
+
 0.16.0 (2021-10-16)
 ~~~~~~~~~~~~~~~~~~~
 - Add support for Python 3.10
--- a/pytest_asyncio/plugin.py
+++ b/pytest_asyncio/plugin.py
@@ -80,8 +80,11 @@ class FixtureStripper:
 def pytest_fixture_post_finalizer(fixturedef, request):
     """Called after fixture teardown"""
     if fixturedef.argname == "event_loop":
-        # Set empty loop policy, so that subsequent get_event_loop() provides a new loop
-        asyncio.set_event_loop_policy(None)
+        policy = asyncio.get_event_loop_policy()
+        policy.get_event_loop().close()  # Clean up existing loop to avoid ResourceWarnings
+        new_loop = policy.new_event_loop()  # Replace existing event loop
+        # Ensure subsequent calls to get_event_loop() succeed
+        policy.set_event_loop(new_loop)
 
 
 @pytest.hookimpl(hookwrapper=True)
--- a/tests/respect_event_loop_policy/conftest.py
+++ b/tests/respect_event_loop_policy/conftest.py
@@ -1,0 +1,16 @@
+"""Defines and sets a custom event loop policy"""
+import asyncio
+from asyncio import DefaultEventLoopPolicy, SelectorEventLoop
+
+
+class TestEventLoop(SelectorEventLoop):
+    pass
+
+
+class TestEventLoopPolicy(DefaultEventLoopPolicy):
+    def new_event_loop(self):
+        return TestEventLoop()
+
+
+# This statement represents a code which sets a custom event loop policy
+asyncio.set_event_loop_policy(TestEventLoopPolicy())
--- a/tests/respect_event_loop_policy/test_respects_event_loop_policy.py
+++ b/tests/respect_event_loop_policy/test_respects_event_loop_policy.py
@@ -1,0 +1,16 @@
+"""Tests that any externally provided event loop policy remains unaltered."""
+import asyncio
+
+import pytest
+
+
+@pytest.mark.asyncio
+async def test_uses_loop_provided_by_custom_policy():
+    """Asserts that test cases use the event loop provided by the custom event loop policy"""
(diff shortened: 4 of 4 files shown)


seifertm:
@wouterdb @bryevdv Would this work for you?

wouterdb:
I think this will work for us, thank you.

bryevdv:
I'm not really knowledgeable enough to comment, but it seems plausible and we'll certainly try updating versions when it's available.

seifertm:
@Tinche What are your thoughts on this?

seifertm:
ping @Tinche :)

seifertm:
I acknowledge that I created this PR without a previous design discussion. However, I do think this change fixes a real world problem and that the new fixture finalizer is cleaner than the existing code.

@Tinche If you can make the time for it, I'd be happy to adjust and rebase this PR or try a different approach based on your input on the matter.

asvetlov (approved):
LGTM.
A not-started event loop is a cheap object, it has only a pipe for self-wakeup plus a few empty data containers.

seifertm:
Thanks for the feedback! I just rebased to master to fix a conflict in the changelog. It turns out the PR is now causing issues on Python 3.8. I'll have another look later today.

seifertm:
I did the following changes to the PR:
- Adjusted formatting to satisfy black and flake8
- During tear down of the `event_loop` fixture:  Explicitly `close()` the current event loop before requesting a `new_event_loop()` from the loop policy. This fixes an issue on Python 3.8 where random tests would fail due to a `ResourceWarning`. In my case, the warning was about unclosed Unix sockets.

All green now :)

Tinche:
GJ, merging in.

If you're working on event loop teardown, there's an issue laying around here somewhere about shutting down the loop more robustly (https://github.com/pytest-dev/pytest-asyncio/issues/222), think we could incorporate that too?

asvetlov:
Regarding #222 -- I think #235 can fix the proper teardown along with many other things.
I hope I can find time for PR soon.
If anybody can help me -- you are welcome, sure!

asvetlov:
Please don't get me wrong -- I'm not opposed to step-by-step improvements.
I just see a generic solution for very many challenges; I wrote `aioloop-proxy` keeping in mind both `pytest-asyncio` and `unittest.AsyncioTestCase`.

Issue #191, opened 2020-10-05
Drop support for Python 3.5


seifertm:
Python 3.5 [reached end-of-life](https://www.python.org/dev/peps/pep-0478/#release-schedule) on 2020-09-30. We should consider dropping support for it.

I would happily take care of it, either in a separate PR or as part of #190 .

Tinche:
I agree, we can drop 3.5. Could you put together a PR and I can merge that in too. Thanks!

seifertm:
Python 3.5 support was removed as of v0.15

PR #193, opened 2020-10-06
Remove python3.5 support

seifertm:
Adresses #191.

This PR removes support for Python 3.5 and consists of three commits:
1. Actual removal of Python 3.5 support, corresponding CI pipelines, documentation, and the dependency on the `async_generator` library.
2. A refactoring commit that deduplicates test cases in the modules `test_simple_35` and `test_simple`
3. A refactoring commit that removes the `_35` and `_36` suffixes from test module names. Test modules with these suffixes give the impression that they are associated with a certain Python version. This is no longer the case.


README.rst
```diff
@@ -178,6 +165,7 @@ Changelog
 0.15.0 (UNRELEASED)
 ~~~~~~~~~~~~~~~~~~~
 - Add support for Python 3.9
+- Abandon support for Python 3.5
```
Tinche, line 168: Just add a note here instructing users to use a lower version if they need 3.5 support :)

Tinche (changes requested):

README.rst
seifertm: Sure! I adjusted the Changelog accordingly. Let me know what you think.

Tinche:
Cool, thanks a lot!

PR #216, opened 2021-06-08
Add timeout options

### LINKED ISSUE #215: Test timeout via marker kwargs
At [dask.distributed](https://github.com/dask/distributed) we started using `pytest-asyncio` for increasingly many tests. Before this, we've been decorating our plain asyncio tests using our own decorator which is not unlike to what `pytest-asyncio` is using under the hood, see https://github.com/dask/distributed/blob/833c5f6c040feaa4550fa343d8e6e4feef3f84d5/distributed/utils_test.py#L753-L772 but that has the obvious shortcoming that it doesn't integrate that nicely into `pytest`.

For our use case the timeout argument is very important since many tests are waiting until certain conditions are true. Moving to `pytest-asyncio` required us to use https://pypi.org/project/pytest-timeout/ but for portability reasons this plugin terminates the entire test process if a single test timeouts (see thread vs signal in their docs) which is not great. 

A better integration with asyncio would be relatively easy to achieve and would allow us to improve our integration with pytest further. The syntax I had in mind is simply

```python
@pytest.mark.asyncio(timeout=30)
async def test_requires_timeout():
    await stuff()
```

which, under the hood, would wrap the coro with https://docs.python.org/3/library/asyncio-task.html#asyncio.wait_for and would raise an `asyncio.TimeoutError` if the test runs for too long.
A default value for this timeout could be provided via a configuration file or similar.

If this overlap in scope is acceptable for this repo I would prepare and open a PR.

fjetter:
This adds the option to add timeouts to tests either via marker kwargs or via options/CLI arguments.

The naming of the options is intentionally prefixes with asyncio to avoid clashes with `pytest-timeout`. Marker kwargs do not have this ambiguity and I therefore allowed both.

I don't have a good sense on how to test the options via config and CLI. If somebody got an idea, I'd be happy to implement it.

Closes #215 
Depends on pytest-dev/pytest-timeout#117


asvetlov:
Please read https://docs.pytest.org/en/6.2.x/writing_plugins.html#testing-plugins

`pytester` fixture can be used for testing command-line and `ini` files.

The plugin has no documentation, IMHO.
I can suggest only the [source code](https://docs.pytest.org/en/6.2.x/_modules/_pytest/pytester.html) and [example of usage in aiohttp](https://github.com/aio-libs/aiohttp/blob/master/tests/test_pytest_plugin.py#L140). 

Here is an example of calling pytest with custom cmdline arguments: https://github.com/aio-libs/aiohttp/blob/master/tests/test_pytest_plugin.py#L140-L150

fjetter:
Didn't know about `pytester`. Thanks for adding the commits!

asvetlov:
The PR is not complete until timeout disabling on debugging is implemented, as in pytest-timeout.
Otherwise, any tests debug session ends with TimeoutError quickly.
I'm not sure is it worth to copy-paste the whole implementation from pytest-timeout plugin?
Maybe adding timeout_setup()/timeout_teardown() hooks for implementing them in pytest-asyncio makes the code easier and stablier?
Should take time for experimenting.

altendky:
I'm interested in timeouts for my asyncio tests and it looks like this might be the path to that.  Is that correct?  What is needed here?

johnthagen:
Any chance that this PR could make progress? This seems like a really important fundamental tool for testing `async` functions.

PR #236, opened 2022-01-05
Remove use of obsolete transfer_markers during test collection phase

seifertm:
`transfer_markers` was removed in Pytest 4.1. The minimum required pytest version for pytest-asyncio is currently v5.4.0. That means that `transfer_markers` is always a no-op.


Tinche:
Thanks!

PR #237, opened 2022-01-06
Drop support for Python 3.6

seifertm:
CPython 3.6 reached end-of-life on 2021-12-23 [1]. Future releases of pytest will no longer support Python 3.6 [2].

We should consider dropping support for unmaintained Python versions.

[1] https://www.python.org/dev/peps/pep-0494/
[2] https://github.com/pytest-dev/pytest/pull/9437


asvetlov (approved):
Agree.
LGTM, thanks for the pull request!

Tinche:
@seifertm thanks!

PR #220, opened 2021-07-10
Resolve event loop closed issue when trying a second attempt with `flaky`

wjsi:
When using [flaky](https://github.com/box/flaky) along with `pytest-asyncio`, for instance,

```python
@flaky.flaky(3)
@pytest.mark.asyncio
async def test_errors():
    raise ValueError
```

when the first attempt failed, the second attempt will raise `RuntimeError: Event loop is closed` (given https://github.com/box/flaky/issues/166). The reason is that the wrapper https://github.com/pytest-dev/pytest-asyncio/blob/6ec76477061ea14394cadbf2cef673b04971ef4d/pytest_asyncio/plugin.py#L185 will be called twice in  `pytest_pyfunc_call` and the previous closed loop is still used. The resolution is to record the original function and wrap the original function every time.

Fixes #178

PR #239, opened 2022-01-06
Prevent double wrapping of inherited Hypothesis tests

### LINKED ISSUE #231: Shared tests with hypothesis throws an Event loop is closed error
When you have a class containing some shared tests (e.g. for contract testing between several implementations of an interface), and use both hypothesis and pytest-asyncio on the shared tests, pytest will raise an error when running the test for the second time.

https://github.com/HypothesisWorks/hypothesis/issues/3122

### Example
```python
import hypothesis
import hypothesis.strategies
import pytest

class Shared:
    @pytest.mark.asyncio
    @hypothesis.given(value=hypothesis.strategies.integers())
    async def test_foo(self, value: int) -> None:
        assert value == value

class TestCaseOne(Shared):
    pass

class TestCaseTwo(Shared):
    pass
```

```
================================= test session starts ==================================
platform darwin -- Python 3.9.4, pytest-6.2.5, py-1.10.0, pluggy-1.0.0
rootdir: /Users/garyevans/Projects/pytest-async-hypothesis
plugins: asyncio-0.16.0, hypothesis-6.23.3
collected 2 items                                                                      

test.py .F                                                                       [100%]

======================================= FAILURES =======================================
_________________________________ TestCaseTwo.test_foo _________________________________

self = <test.TestCaseTwo object at 0x1069abe20>

    @pytest.mark.asyncio
>   @hypothesis.given(value=hypothesis.strategies.integers())

test.py:8: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
../../Library/Caches/pypoetry/virtualenvs/pytest-async-hypothesis-BRvTXcHz-py3.9/lib/python3.9/site-packages/pytest_asyncio/plugin.py:191: in inner
    coro = func(**kwargs)
../../Library/Caches/pypoetry/virtualenvs/pytest-async-hypothesis-BRvTXcHz-py3.9/lib/python3.9/site-packages/pytest_asyncio/plugin.py:193: in inner
    task = asyncio.ensure_future(coro, loop=_loop)
../../.pyenv/versions/3.9.4/lib/python3.9/asyncio/tasks.py:667: in ensure_future
    task = loop.create_task(coro_or_future)
../../.pyenv/versions/3.9.4/lib/python3.9/asyncio/base_events.py:431: in create_task
    self._check_closed()
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <_UnixSelectorEventLoop running=False closed=True debug=False>

    def _check_closed(self):
        if self._closed:
>           raise RuntimeError('Event loop is closed')
E           RuntimeError: Event loop is closed

../../.pyenv/versions/3.9.4/lib/python3.9/asyncio/base_events.py:510: RuntimeError
-------------------------------------- Hypothesis --------------------------------------
Falsifying example: test_foo(
    value=0, self=<test.TestCaseTwo at 0x1069abe20>,
)
=============================== short test summary info ================================
FAILED test.py::TestCaseTwo::test_foo - RuntimeError: Event loop is closed
============================= 1 failed, 1 passed in 0.13s ==============================
sys:1: RuntimeWarning: coroutine 'Shared.test_foo' was never awaited
```
### Environment
```
Package          Version
---------------- -------
hypothesis       6.23.3
pytest           6.2.5
pytest-asyncio   0.16.0
```

### Workaround
If we modify the `event_loop` fixture and do not close the loop by default then the tests pass as expected.
https://github.com/pytest-dev/pytest-asyncio/blob/f21e0da345f877755b89ff87b6dcea70815b4497/tests/test_hypothesis_integration.py#L11-L15

seifertm:
When looking at #231 I thought that something is wrong with the setup and teardown of the `event_loop` fixture. Running pytest with `--show-setup` proved that this is not the case, though:

```
tests/hypothesis/test_inherited_test.py 
        SETUP    F event_loop
        tests/hypothesis/test_inherited_test.py::TestOne::test_hypothesis (fixtures used: event_loop).
        TEARDOWN F event_loop
        SETUP    F event_loop
        tests/hypothesis/test_inherited_test.py::TestTwo::test_hypothesis (fixtures used: event_loop).
        TEARDOWN F event_loop
```

I dug further and noticed that regular async tests are executed just fine. Only tests that were using Hypothesis raised errors when they were executed in a subclass. Eventually, it turned out that subclassing caused multiple synchronous function wrappers to be layered around inherited Hypothesis tests. Here's why:

Pytest-asyncio identifies Hypothesis test cases by their `is_hypothesis_test` attribute. When setting up an async Hypothesis test pytest-asyncio replaces the function's `hypothesis.inner_test` attribute. The the top level function never changes.
    
When a Hypothesis test case is defined in a base class and inherited by subclasses, the test is collected in each subclass. Since the top-level Hypothesis test never changes, its inner test will be wrapped multiple times.
    
Double wrapping leads to execution errors caused by stale (closed) event loops in all test executions after the first. This also explains why a module scoped fixture could serve as a workaround for the OP in #231.

This change adds an `original_test_function` attribute to the async function wrapper, in order to keep track of the original Hypothesis test. When re-wrapping would occur in subclasses pytest-asyncio wraps the original test function rather than the wrapper function.

Closes #231


seifertm:
Rebased to master. #220 addresses the same issue in a slightly different way. I adjusted the PR accordingly.

The PR now:
* Fixes a linting issue
* Links the fixes issues in the changelog
* Mentions flaky as an added test dependency in the changelog. This is an important information for downstream packagers.

asvetlov (approved):
LGTM

PR #240, opened 2022-01-07
Reformat with black

asvetlov:
`setup.py` is also reformatted for the sake of uniformness.


Tinche (approved):

Tinche:
@asvetlov LGTM. As for the workflow, would you like me to just approve and leave the merging to you, or merge too?

asvetlov:
Up to you.
In CPython, we usually prefer to give the PR author to decide when pressing the merge button if the author is a core developer.
The rule is not strict though, sometimes we can violate the rule if the feature is desirable to land quickly.
Regarding self requests, there is another rule: wait for a day at least if nobody the PR reviewed yet. Again, the rule is soft.

I'm pretty sure other mature Open Source projects have a similar practice.

PR #252, opened 2022-01-12
Release process automation

asvetlov:
The release is created on git tag pushing.

After linters and tests passed, the deploy stage is starting to wait for release approval.

@pytest-dev/pytest-asyncio-admin can approve. If nobody approved the release in 1 day -- deployment is discarded.
After the approval, we have 15 minutes timer for starting the deployment. During these 15 mins we the deployment task can be cancelled (just in case the deployer missed something).

The new release is pushed on PyPI (using generated PyPI upload token with `pytest-asyncio` scope) and on GitHub (visible as GitHub release).

Tag naming: as usual.
`v0.16`, `v0.17`  etc means final releases.
`v0.17a0` means *0.17 alpha 0* release.
`.post` releases are supported but please avoid them -- Dependabot uses SemVer; SemVer handles it as pre-release. Better to publish the next micro-release instead.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -17,11 +17,12 @@ jobs:
       prerelease: ${{ steps.version.outputs.prerelease }}
     steps:
       - uses: "actions/checkout@v2"
+        fetch-depth: 0
       - uses: "actions/setup-python@v2"
         with:
           python-version: "3.9"
 
-      - name: "Install poetry, check-wheel-content, and twine"
+      - name: "Install check-wheel-content, and twine"
         run: "python -m pip install build check-wheel-contents tox twine"
       - name: "Build package"
         run: "python -m build"
@@ -53,6 +54,7 @@ jobs:
 
     steps:
       - uses: "actions/checkout@v2"
+        fetch-depth: 0
       - uses: "actions/setup-python@v2"
         with:
           python-version: "${{ matrix.python-version }}"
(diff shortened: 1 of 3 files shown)


.github/workflows/main.yml
```diff
@@ -9,6 +9,38 @@ on:
   workflow_dispatch:
 
 jobs:
+  lint:
```
seifertm, line 12: The logs for the *Build package* step shows the name of the built artifacts is `pytest-asyncio-0.1.dev1+ga5205f2.tar.gz` and `pytest_asyncio-0.1.dev1+ga5205f2-py3-none-any.whl`. As far as I understand these are reused in the `deploy` stage later.

I expected them to be to have a version like `pytest_asyncio-0.16.1+gabcd1234…` not `0.1.dev1`.

The checkout action seems to do a shallow clone with `--no-tags`. Is it possible this prevents setuptools_scm from determining the correct version?

.github/workflows/main.yml
```diff
@@ -9,6 +9,38 @@ on:
...
+    outputs:
+      version: ${{ steps.version.outputs.version }}
+      prerelease: ${{ steps.version.outputs.prerelease }}
+    steps:
+      - uses: "actions/checkout@v2"
+      - uses: "actions/setup-python@v2"
+        with:
+          python-version: "3.9"
+
+      - name: "Install poetry, check-wheel-content, and twine"
```
seifertm, line 24: The Poetry part in the description no longer seems to be valid. Could you adjust it?

Tinche (changes requested):
Hm it install poetry in the job?

.github/workflows/main.yml
```diff
@@ -51,23 +83,32 @@ jobs:
...
-        run: "python -m build"
-      - name: "List result"
-        run: "ls -l dist"
-      - name: "Check wheel contents"
-        run: "check-wheel-contents dist/*.whl"
-      - name: "Check long_description"
-        run: "python -m twine check dist/*"
+    - name: Checkout
+      uses: actions/checkout@v2.4.0
+    - name: Download disctributions
```
nicoddemus, line 94:
```suggestion
    - name: Download distributions
```

asvetlov:
@Tinche we had it in the workflow previously.
Fixed now.

.github/workflows/main.yml
asvetlov: `fetch-depth: 0` should fix it. Let's see

eirnym:
@asvetlov nice algorithm! I'd like to apply similar to my projects. However, I don't see how timers and confirmations are set in configuration. Could you enlighten me?

asvetlov:
@eirnym  I cannot show you our Settings config, it is available for the project admin only.

Please take a look at [GitHub docs first](https://docs.github.com/en/actions/deployment/targeting-different-environments/using-environments-for-deployment) and feel free to ask if you have a question.

asvetlov:
Merging, I'd like to build an alpha release to test the workflow

PR #254, opened 2022-01-13
Alternative yamlfmt settings

asvetlov:
Default `yamlfmt` style is wider than I usually use.

The proposed PR suggests a much denser format.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,14 +1,14 @@
 ---
 version: 2
 updates:
-    - package-ecosystem: pip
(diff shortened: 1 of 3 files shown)


Tinche (approved):

asvetlov:
@Tinche @seifertm what do you think?
I can live with any auto-formatting, it is better than no formatting style at all.
But I think this version is slightly better readable (and looks like yaml formatting from GitHub docs).

Tinche:
Yeah, I can see how the original was too wide.

PR #259, opened 2022-01-13
Fix auto marking of async hypothesis tests in auto mode

### LINKED ISSUE #258: asyncio_mode=auto does not support hypothesis without marker
The new `asyncio_mode=auto` mode does not support hypothesis markers. I'm assuming this is an issue on the asyncio side, since `strict` mode continues to work, but please let me know if I should post this to `hypothesis` instead!

I think @asvetlov and @seifertm are the relevant folks to ping for this.

Reproduction case:

<details>
<summary>pytest.ini</summary>

```
[pytest]
asyncio_mode = auto
```
</details>

<details>
<summary>test.py</summary>

```
from hypothesis import given
from hypothesis import strategies as st

@given(x=st.booleans())
async def test_foo(x):
    assert x
```
</details>

```
» pip install pytest-asyncio hypothesis
» pytest test.py
== test session starts =
platform darwin -- Python 3.7.12, pytest-6.2.5, py-1.11.0, pluggy-1.0.0
rootdir: /private/tmp/testpyasynchypo, configfile: pytest.ini
plugins: hypothesis-6.35.0, asyncio-0.17.0
collected 1 item                                                                                                                                                                                                            

test.py F                                                                       [100%]

== FAILURES ==
_ test_foo _

    @given(x=st.booleans())
>   async def test_foo(x):
E   hypothesis.errors.InvalidArgument: Hypothesis doesn't know how to run async test functions like test_foo.  You'll need to write a custom executor, or use a library like pytest-asyncio or pytest-trio which can handle the translation for you.
E       See https://hypothesis.readthedocs.io/en/latest/details.html#custom-function-execution

test.py:6: InvalidArgument
== short test summary info ==
FAILED test.py::test_foo - hypothesis.errors.InvalidArgument: Hypothesis doesn't know how to run async test functions like test_foo.  You'll need to write a custom executor, or use a library like pytest-asyncio or pyte...
== 1 failed in 0.22s ==
```

Comparison, using `strict` mode, the following works as expected:

<details>
<summary>pytest.ini</summary>

```
[pytest]
asyncio_mode = strict
```
</details>

<details>
<summary>test.py</summary>

```
import pytest
from hypothesis import given
from hypothesis import strategies as st

@pytest.mark.asyncio
@given(x=st.booleans())
async def test_foo(x):
    assert x
```
</details>

seifertm:
The option `--asyncio-mode=auto` marks all async functions with the `asyncio` mark during the collection phase. However, when pytest collects the Hypothesis test, the `@given` decorator has already been applied and the Hypothesis test function is no longer a coroutine.
    
This commit extends the `pytest_pycollect_makeitem` hook to mark Hypothesis tests whose function body is a coroutine.
    
 Closes #258
 

I was uncertain whether to add  the tests to the auto mode tests or to the Hypothesis tests. I eventually decided to add them to the tests for the Hypothesis integration. Rather than using `pytester` I added some assertions so the tests would fail in case we ever change `asyncio_mode` to something other than auto. Let me know what you think of this.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -256,6 +256,10 @@ or an async framework such as `asynctest <https://asynctest.readthedocs.io/en/la
 
 Changelog
 ---------
+0.17.1 (UNRELEASED)
+~~~~~~~~~~~~~~~~~~~
+- Fixes a bug that prevents async Hypothesis tests from working without explicit ``asyncio`` marker when ``--asyncio-mode=auto`` is set. `#258 <https://github.com/pytest-dev/pytest-asyncio/issues/258>`_
+
 0.17.0 (22-01-13)
 ~~~~~~~~~~~~~~~~~~~
 - `pytest-asyncio` no longer alters existing event loop policies. `#168 <https://github.com/pytest-dev/pytest-asyncio/issues/168>`_, `#188 <https://github.com/pytest-dev/pytest-asyncio/issues/168>`_
--- a/pytest_asyncio/plugin.py
+++ b/pytest_asyncio/plugin.py
@@ -115,7 +115,13 @@ def pytest_configure(config):
 @pytest.mark.tryfirst
 def pytest_pycollect_makeitem(collector, name, obj):
     """A pytest hook to collect asyncio coroutines."""
-    if collector.funcnamefilter(name) and _is_coroutine(obj):
+    if not collector.funcnamefilter(name):
+        return
+    if (
+        _is_coroutine(obj)
+        or _is_hypothesis_test(obj)
+        and _hypothesis_test_wraps_coroutine(obj)
+    ):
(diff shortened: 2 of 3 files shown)


Tinche (approved):

asvetlov (approved):
I can live with it, sure.
The only note: when we change `asyncio_mode` to strict in `setup.cfg` these tests will fail.
That's why I personally prefer `pytester`.

seifertm:
You're right. It's probably a better idea if the tests are decoupled from the project config. I was hoping to avoid the boilerplate that comes with `pytester`, but maybe we'll find another solution in the future :)

I rewrote the two test cases to use `pytester` which runs the tests with `--asyncio-mode=auto`.

PR #260, opened 2022-01-14
Provide typing info

### LINKED ISSUE #198: Add type annotations

asvetlov:
Type checking is rough, better checks require not released yet pytest 7.0 at least.

Tests are not typed to don't pollute this big enough PR.
They can be typed later.

Fixes #198

### DIFF
--- a/pytest_asyncio/plugin.py
+++ b/pytest_asyncio/plugin.py
@@ -6,8 +6,36 @@ import functools
 import inspect
 import socket
 import warnings
+from typing import (
+    Any,
+    Awaitable,
+    Callable,
+    Dict,
+    Iterable,
+    Iterator,
+    List,
+    Optional,
+    Set,
+    TypeVar,
+    Union,
+    cast,
+    overload,
+)
 
 import pytest
+from typing_extensions import Literal
+
+_ScopeName = Literal["session", "package", "module", "class", "function"]
+_T = TypeVar("_T")
+
+FixtureFunction = TypeVar("FixtureFunction", bound=Callable[..., object])
+FixtureFunctionMarker = Callable[[FixtureFunction], FixtureFunction]
+
+Config = Any  # pytest < 7.0
+PytestPluginManager = Any  # pytest < 7.0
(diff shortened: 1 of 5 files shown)


pytest_asyncio/plugin.py
```diff
@@ -6,8 +6,36 @@
...
+    overload,
+)
 
 import pytest
+from typing_extensions import Literal
+
+_ScopeName = Literal["session", "package", "module", "class", "function"]
+_T = TypeVar("_T")
+
+FixtureFunction = TypeVar("FixtureFunction", bound=Callable[..., object])
```
Tinche, line 31: This is supposed to only be applied to async functions, right?

It should probably look like this: https://github.com/hynek/prometheus-async/pull/21/files#diff-df328b684166777b1785c7abc3acfcb1c1ec5c236567bc39472d1389726267a7R32

Maybe the ParamSpec is unnecessary, but we can verify the return type is a subtype of Awaitable.

Tinche (changes requested):

pytest_asyncio/plugin.py
asvetlov: Good idea.
Actually, it can be either an Awaitable or AsyncIterator.
I thought about it but was too lazy.

Tinche: I understand completely. ;)

asvetlov: Done.
`ParamSpec` cannot be applied as `Callable[ParamSpec, RetType]` argument in `mypy==0.931`.
The `ParamSpec` implementation is not full yet, unfortunately. I hope it will be fixed sooner or later, ParamSpec looks like the very useful thing.

asvetlov: tricky

asvetlov:
The PR is green

asvetlov:
Merged, if we require the type info -- better to add it fast, other PRs can use typing from the very beginning.

jeremad:
That's great thank you very much. Sorry for the lack of response to your question in the first place.

PR #261, opened 2022-01-14
Don't close event loop if the loop doesn't exist

### LINKED ISSUE #257: Event loop shutdown failures on 0.17
We use the following to define a session scoped event loop (per https://github.com/pytest-dev/pytest-asyncio/issues/68)

```python
@pytest.fixture(scope="session")
def event_loop(request):
    """
    Redefine the event loop to support session/module-scoped fixtures;
    see https://github.com/pytest-dev/pytest-asyncio/issues/68
    """
    loop = policy.new_event_loop()

    try:
        yield loop
    finally:
        loop.close()
```

After upgrading to 0.17, our test suite fails with

```python
_______________________________________________________________________________________ ERROR at teardown of test_XXXX_______________________________________________________________________________________

fixturedef = <FixtureDef argname='event_loop' scope='session' baseid='tests'>, request = <SubRequest 'event_loop' for <Function test_secret_settings_are_not_serialized>>

    @pytest.hookimpl(trylast=True)
    def pytest_fixture_post_finalizer(fixturedef, request):
        """Called after fixture teardown"""
        if fixturedef.argname == "event_loop":
            # Clean up existing loop to avoid ResourceWarnings
>           policy.get_event_loop().close()

/opt/homebrew/Caskroom/miniconda/base/envs/orion-dev-38/lib/python3.8/site-packages/pytest_asyncio/plugin.py:163: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <asyncio.unix_events._UnixDefaultEventLoopPolicy object at 0x7ff8eaa8f8b0>

    def get_event_loop(self):
        """Get the event loop for the current context.
    
        Returns an instance of EventLoop or raises an exception.
        """
        if (self._local._loop is None and
                not self._local._set_called and
                isinstance(threading.current_thread(), threading._MainThread)):
            self.set_event_loop(self.new_event_loop())
    
        if self._local._loop is None:
>           raise RuntimeError('There is no current event loop in thread %r.'
                               % threading.current_thread().name)
E           RuntimeError: There is no current event loop in thread 'MainThread'.

/opt/homebrew/Caskroom/miniconda/base/envs/orion-dev-38/lib/python3.8/asyncio/events.py:639: RuntimeError
```

Modifying the fixture to set the loop with the policy at the end resolves the error

```python
    try:
        yield loop
    finally:
        loop.close()

    policy.set_event_loop(loop)
```

Leaving the loop open or setting the loop with the policy before yielding does not resolve the error.

I'm not sure why this is happening now, not sure if it needs a fix here.

We're using `asyncio_mode = auto`.

asvetlov:
Fixes #257

### DIFF
--- a/pytest_asyncio/plugin.py
+++ b/pytest_asyncio/plugin.py
@@ -169,8 +169,13 @@ def pytest_fixture_post_finalizer(fixturedef, request):
     """Called after fixture teardown"""
     if fixturedef.argname == "event_loop":
         policy = asyncio.get_event_loop_policy()
-        # Clean up existing loop to avoid ResourceWarnings
-        policy.get_event_loop().close()
+        try:
+            loop = policy.get_event_loop()
+        except RuntimeError:
+            loop = None
+        if loop is not None:
+            # Clean up existing loop to avoid ResourceWarnings
+            loop.close()
         new_loop = policy.new_event_loop()  # Replace existing event loop
         # Ensure subsequent calls to get_event_loop() succeed
         policy.set_event_loop(new_loop)
--- a/README.rst
+++ b/README.rst
@@ -259,6 +259,7 @@ Changelog
 0.17.1 (UNRELEASED)
 ~~~~~~~~~~~~~~~~~~~
 - Fixes a bug that prevents async Hypothesis tests from working without explicit ``asyncio`` marker when ``--asyncio-mode=auto`` is set. `#258 <https://github.com/pytest-dev/pytest-asyncio/issues/258>`_
+- Fixed a bug that closes the default event loop if the loop doesn't exist `#258 <https://github.com/pytest-dev/pytest-asyncio/issues/257>`_
 
 0.17.0 (22-01-13)
 ~~~~~~~~~~~~~~~~~~~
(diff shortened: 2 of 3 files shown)


asvetlov:
Ready for review, tests are green

pytest_asyncio/plugin.py
```diff
@@ -169,8 +169,13 @@ def pytest_fixture_post_finalizer(fixturedef, request):
     """Called after fixture teardown"""
     if fixturedef.argname == "event_loop":
         policy = asyncio.get_event_loop_policy()
-        # Clean up existing loop to avoid ResourceWarnings
-        policy.get_event_loop().close()
+        try:
+            loop = policy.get_event_loop()
+        except RuntimeError:
+            loop = None
```
seifertm, line 175: Would it make sense to emit a warning here? After all, if `get_event_loop` fails it means that someone was messing with the loop provided by the `event_loop` fixture. A warning would at least tell the users they are going down a dangerous path.

seifertm (approved):

madkinsz:
Thanks! This implementation lgtm

PR #269, opened 2022-01-17
Require typing-extensions on py<3.8 only

mgorny:
typing.Literal is available since Python 3.8.  Use it instead of
requiring typing-extensions for systems that no longer have Python 3.7.

### DIFF
--- a/pytest_asyncio/plugin.py
+++ b/pytest_asyncio/plugin.py
@@ -5,6 +5,7 @@ import enum
 import functools
 import inspect
 import socket
+import sys
 import warnings
 from typing import (
     Any,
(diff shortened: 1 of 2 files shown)


pytest_asyncio/plugin.py
```diff
@@ -24,7 +25,11 @@
 )
 
 import pytest
-from typing_extensions import Literal
+
+if sys.version_info >= (3, 8):
```
mgorny, line 29: I can change this to use `sys.hexversion` or `ImportError` approach if you prefer.

Tinche (approved):

PR #270, opened 2022-01-17
Code cleanup: drop 'request' argument from async fixtures

asvetlov:
The fixture injection is not needed, pytest 6.1.0+ works fine with using `request` argument from `pytest_fixture_setup()` hook.


seifertm (approved):
Good find!

asvetlov:
@seifertm thanks!
To be clear, `fixtredef.addfinalizer()` does the same work. Actually, `request.addfinalizer()` calls `fixturedef.addfinalizer()` internally.

Issue #271, opened 2022-01-18
0.17.* issues with tornado loop started by dask distributed


bphillips-exos:
Hello, the following test fails on teardown. I believe this is due to the tornado IOLoop that is started when the cluster is initialized. This example works on 0.16.0 and fails when upgrading to 0.17.0+. 

```python3
from dask.distributed import LocalCluster

@pytest.mark.asyncio
async def test_dask() -> None:
    cluster = LocalCluster()
```

```
tests/compute/test_cluster.py:20 (test_dask)
fixturedef = <FixtureDef argname='event_loop' scope='function' baseid=''>
request = <SubRequest 'event_loop' for <Function test_dask>>

    @pytest.hookimpl(trylast=True)
    def pytest_fixture_post_finalizer(fixturedef: FixtureDef, request: SubRequest) -> None:
        """Called after fixture teardown"""
        if fixturedef.argname == "event_loop":
            policy = asyncio.get_event_loop_policy()
            try:
                loop = policy.get_event_loop()
            except RuntimeError:
                loop = None
            if loop is not None:
                # Clean up existing loop to avoid ResourceWarnings
>               loop.close()

.venv/lib/python3.9/site-packages/pytest_asyncio/plugin.py:269: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/asyncio/unix_events.py:58: in close
    super().close()
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <_UnixSelectorEventLoop running=True closed=False debug=False>

    def close(self):
        if self.is_running():
>           raise RuntimeError("Cannot close a running event loop")
E           RuntimeError: Cannot close a running event loop

/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/asyncio/selector_events.py:89: RuntimeError
```

Versions: (reproduced on multiple versions of dask / distributed)
* dask 2022.1.0
* distributed 2022.1.0
* pytest 6.2.5
* pytest-asyncio 0.17.2

bphillips-exos:
[...]

PR #272, opened 2022-01-18
Only close the loop if it is not running

bphillips-exos:
Resolves https://github.com/pytest-dev/pytest-asyncio/issues/271


asvetlov:
I think that the `session` scoped `event_loop` fixture should be used instead.
A silent suppressing of the loop closing looks dangerous.
As an option, we can raise a warning at least (for the sake of keeping backward compatibility with 0.16).

PR #275, opened 2022-01-21
Raise a warning if @pytest.mark.asyncio is applied to non-async function

asvetlov:


pytest_asyncio/plugin.py
```diff
@@ -424,34 +432,43 @@ def wrap_in_sync(func: Callable[..., Awaitable[Any]], _loop: asyncio.AbstractEve
...
-                if task.done() and not task.cancelled():
-                    task.exception()
-                raise
+        if not inspect.isawaitable(coro):
+            pyfuncitem.warn(
+                pytest.PytestWarning(
+                    f"The test {pyfuncitem} is marked with '@pytest.mark.asyncio' "
+                    "but it is not an async function. "
+                    "Please remove asyncio marker, "
+                    "use 'asyncio_mode = auto' for autocollection."
```
seifertm, line 441: It makes a lot of sense to issue a warning when the user marks a synchronous function as `asyncio`.

As a user, I would be confused by the instruction to use `asyncio_mode=auto`, though. What does that have to do with the misuse of `pytest.mark.asyncio`?

asvetlov: I'm not sure about the best error text. Please feel free to correct it.
My thought was that the global mark (`pytestmark = pytest.mark.asyncio`) can lead to false markers easily.
Maybe we don't need to emphasize it. `"Please remove asyncio marker."` without mention `asyncio_mode = auto` is an option.
What do you think?

seifertm: Ah, of course! The mark can also be applied when the user writes `pytestmark = pytest.mark.asyncio`. I didn't think of that.

I would definitely mention pytestmark, but leave out `asyncio_mode`. Maybe something like this?

"The test {pyfuncitem} is marked with '@pytest.mark.asyncio'  but it is not an async function. "
"Please remove the asyncio marker. If the test is not marked explicitly, check for global markers applied via 'pytestmark'"

asvetlov: Sounds good to me. Updated the text.

seifertm (approved):

PR #278, opened 2022-01-25
Support parametrized event_loop fixture

asvetlov:
Assume we have a parametrized `event_loop` fixture that provides both `asyncio` and `uvloop` event loops:
```
@pytest.fixture(scope="module", params=[asyncio.new_event_loop, uvloop.new_event_loop])
def event_loop(request):
    loop = request.param()
    yield loop
    loop.close()
```

The PR fixes async fixtures creation to support such parametrization.
If shifts an async fixture function wrapping from fixture setup stage to tests collection stage.
`pytest` has no dedicated fixture collection hook (yet), the PR collects fixtures in `pytest_pycollect_makeitem` hooks.
The native pytest's fixtures manage uses a similar approach internally.

~As a side effect, tests execution is slightly faster. Async-sync wrapper is installed once per fixture definition, not on every fixture call. The boost is visible for non-function scope fixtures only though.~ Actually no, performance is roughly the same.
The crucial point is adding `event_loop` and `request` dependencies *before* `pytest_generate_tests` is called. Otherwise, parametrization can be broken.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -261,6 +261,7 @@ Changelog
 ~~~~~~~~~~~~~~~~~~~
 
 - Raise a warning if @pytest.mark.asyncio is applied to non-async function. `#275 <https://github.com/pytest-dev/pytest-asyncio/issues/275>`_
+- Support parametrized ``event_loop`` fixture. `#278 <https://github.com/pytest-dev/pytest-asyncio/issues/278>`_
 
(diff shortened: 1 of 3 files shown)


Tinche:
Love this functionality!

Tinche (approved):

PR #279, opened 2022-01-28
Fix setuptools_scm configuration

asvetlov:
`pyproject.toml` should use `setuptools_scm[toml]` dependency.
In turn, bare `setuptools_scm` should be dropped from `setup.cfg` `setup_requires` section.

@webknjaz suggested the change for another aio-libs project.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -2,7 +2,7 @@
 requires = [
   "setuptools>=51.0",
   "wheel>=0.36",
-  "setuptools_scm>=6.2"
+  "setuptools_scm[toml]>=6.2"
 ]
 build-backend = "setuptools.build_meta"
 
--- a/setup.cfg
+++ b/setup.cfg
@@ -34,9 +34,6 @@ python_requires = >=3.7
 packages = find:
 include_package_data = True
 
-setup_requires =
-  setuptools_scm >= 6.2
-
 install_requires =
   pytest >= 6.1.0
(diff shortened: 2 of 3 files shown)


Tinche (approved):

asvetlov:
Interesting, tests are failed.
@webknjaz could you please take a look at what is wrong with the PR?
Should `setup_requires = setuptools_scm>=6.2` be used for the project?

seifertm:
We encountered the same issue in the PR where we integrated `setuptools_scm`. I had a theory why this is happening, but I'm not 100% sure. Cross-reference: https://github.com/pytest-dev/pytest-asyncio/pull/37#issuecomment-1010333770

webknjaz (approved):

webknjaz:
> Should setup_requires = setuptools_scm>=6.2 be used for the project?

This shouldn't be necessary. I can't check right now but my immediate guess would be to check pip/setuptools versions. I'll try to get back to this once I can look at it from my laptop.

webknjaz:
Oh, I see what's happening:
```console
ERROR: FAIL could not package project - v = InvocationError('/opt/hostedtoolcache/Python/3.7.12/x64/bin/python setup.py sdist --formats=zip --dist-dir /home/runner/work/pytest-asyncio/pytest-asyncio/.tox/dist', 1)
```
Tox invokes `setup.py` directly and PEP517 is never used. Enable build isolation to fix this.

PR #284, opened 2022-02-05
Relax dependency on typing-extensions

gazorby:
We only use `typing_extensions` for the `Literal` type, which is available [starting from version `3.7.2`](https://github.com/python/typing/blob/1be6269634bec97d58d68cc1243359118373580a/typing_extensions/src_py3/typing_extensions.py#L129), so we don't necessarily need `typing_extensions>=4.0`

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -36,7 +36,7 @@ include_package_data = True
 
 install_requires =
   pytest >= 6.1.0
-  typing-extensions >= 4.0; python_version < "3.8"
+  typing-extensions >= 3.7.4; python_version < "3.8"
 
 [options.extras_require]
 testing =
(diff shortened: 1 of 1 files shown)


gazorby:
Apparently it was available [starting from version `3.7.2`](https://github.com/python/typing/blob/1be6269634bec97d58d68cc1243359118373580a/typing_extensions/src_py3/typing_extensions.py#L129)

asvetlov (approved):
LGTM
I have a feeling that the PR doesn't make much value (while definitely doesn't harm).
Typing system involves fast, the fresh `typing-extensions` version is crucial for adopting such changes.
Say, I have a plan to use  `ParamSpec` + `Concatenate` once is is fully supported by `mypy`.

Issue #286, opened 2022-02-08
v0.18.0 doesn't find some of my async fixtures


PerchunPak:
[Logs from Github Actions.](https://github.com/PerchunPak/PingerBot/runs/5105657604?check_suite_focus=true#step:7:83)
[Code.](https://github.com/PerchunPak/PingerBot/blob/ec09ca75b3a812de007a97e43628a9fb78fc9cfc/tests/test_other_cmd.py#L9)
Keep in mind, that v0.17.x worked normally.

seifertm:
Thanks for the logs and the code example!
Your code is using `@pytest.fixture` to decorate asynchronous fixtures. Starting with v0.17, asynchronous fixtures should be decorated with `@pytest_asyncio.fixture`. (see https://github.com/pytest-dev/pytest-asyncio#async-fixtures)

Could you try changing the decorators on the fixtures in your test and report back?

PerchunPak:
I am using `@pytest.fixture` in other files, and it works fine with `asyncio_mode = auto`.
Logs when I changed all `@pytest.fixture` to `@pytest_asyncio.fixture` in `test_other_cmd`.
```
================================================= test session starts =================================================
platform win32 -- Python 3.10.2, pytest-7.0.0, pluggy-1.0.0
rootdir: C:\Users\perch\Desktop\programming\PingerBot, configfile: pytest.ini, testpaths: tests
plugins: asyncio-0.18.0, cov-3.0.0, forked-1.3.0, rerunfailures-10.2, timeout-2.0.1, xdist-2.4.0
asyncio: mode=auto
collected 56 items

tests\test_cmd_alias.py ......                                                                                   [ 10%]
tests\test_cmd_motd.py .s....                                                                                    [ 21%]
tests\test_cmd_ping.py ......s.                                                                                  [ 35%]
tests\test_cmd_stat.py ........s....                                                                             [ 58%]
tests\test_database.py ...............                                                                           [ 85%]
tests\test_other_cmd.py FFFFFFFF                                                                                 [100%]

====================================================== FAILURES =======================================================
________________________________________ TestOtherCommands.test_help_commands _________________________________________
tests\test_other_cmd.py:34: in test_help_commands
    for field in help_cmd["fields"]:
E   TypeError: 'coroutine' object is not subscriptable
______________________________________ TestOtherCommands.test_about_description _______________________________________
tests\test_other_cmd.py:40: in test_about_description
    assert bot.app_info.description in about.description
[...]

PR #287, opened 2022-02-08
Fix regression that prevents async fixtures from working in sync tests

seifertm:
The function `_preprocess_async_fixtures` does nothing if the fixture is not an async function. The change simply prepares all async fixtures of a test regardless of whether the test is an async function or not.

Do you see any harm in doing so?

This should fix the issue encountered in #286.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -257,6 +257,10 @@ or an async framework such as `asynctest <https://asynctest.readthedocs.io/en/la
 Changelog
 ---------
 
+0.18.1 (UNRELEASED)
+~~~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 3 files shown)


asvetlov (approved):
LGTM!
Thanks for the fix, my fault.
There are too many combinations between sync and async code :)

asvetlov:
Please let me merge and make a new release right now.
Sorry for the review delay.

seifertm:
> Sorry for the review delay.
@asvetlov No worries. If someone has time to look at these things within a week, I'm already pretty happy :)
You're always very responsive, anyway.

A new release sounds good!

seifertm:
@asvetlov The tag that was being built was `v18.0.1`, instead of `v0.18.1`. I thought that might cause issues and cancelled the running release workflow. Hoping this was okay!

PR #306, opened 2022-03-07
Extract changelog to separate file

seifertm:
This PR extracts the ever-growing changelog to a separate file. The changelog is included in the source tarball for downstream packagers.

Files related to Git, pre-commit and GitHub are removed from the sdist.


asvetlov (approved):
LGTM

Issue #222, opened 2021-07-29
During cleanup, should pytest-asyncio cancel outstanding tasks, shut down async generators, and shut down the thread pool, like `asyncio.run()` does?


SyntaxColoring:
Here's how `pytest-asyncio` implements the `event_loop` fixture:

https://github.com/pytest-dev/pytest-asyncio/blob/6ec76477061ea14394cadbf2cef673b04971ef4d/pytest_asyncio/plugin.py#L224-L229

Note that `loop.close()` is the sole cleanup code.

Compare this with [the implementation of `asyncio.run()`](https://github.com/python/cpython/blob/93d90871d216625b8d9db0fa59693953353d19ae/Lib/asyncio/runners.py#L39-L52):

```python
    loop = events.new_event_loop()
    try:
        events.set_event_loop(loop)
        if debug is not None:
            loop.set_debug(debug)
        return loop.run_until_complete(main)
    finally:
        try:
            _cancel_all_tasks(loop)
            loop.run_until_complete(loop.shutdown_asyncgens())
            loop.run_until_complete(loop.shutdown_default_executor())
        finally:
            events.set_event_loop(None)
            loop.close()
```

Before calling `loop.close()`, `asyncio.run()` does this:

1. It attempts to cancel outstanding tasks with `_cancel_all_tasks()`.
2. It calls `loop.shutdown_asyncgens()`
3. It calls `loop.shutdown_default_executor()`.

(I think (2) and (3) correspond to what [the documentation](https://docs.python.org/3/library/asyncio-task.html#asyncio.run) calls "finalizing asynchronous generators and closing the threadpool.")

I only have a tenuous grasp on these `asyncio` internals, but **it seems like the `event_loop` fixture should do this, too?** And so should any other place in `pytest-asyncio` where the event loop is shut down, if there are more.

andyljones:
Echoing this - I just spent a while figuring out where the heck these 'async generator not closed' warnings were coming from in my test suite, when I couldn't reproduce the behaviour in prod.

podusowski:
+1

Got a bunch of "Task was destroyed but it is pending!" because tasks weren't allowed to handle cancellation and finish.

Tinche:
Yeah, this seems like a good change to make. I'll take a PR or implement this myself in the next couple of weeks.

asvetlov:
Can be solved as a side-effect of https://github.com/pytest-dev/pytest-asyncio/issues/235

Issue #235, opened 2022-01-03
Use a loop proxy for better flexibility


asvetlov:
[aioloop-proxy](https://github.com/aio-libs/aioloop-proxy) provide a proxy instance that is fully compatible with `asyncio.AbstractEventLoop` but allows to cooperate of multiple nested proxies at the same time.

From pytest-asyncio point of view, a session-scope fixture can use a session-level loop proxy that can cooperate with module-level, class-level, and function-level ones providing good test isolation.

I have no proof of concept for pytest-asyncio yet, if somebody with deep pytest internals knowledge can help -- you are welcome!

seifertm:
This sounds interesting! As you mentioned in another comment the loop proxy could address many things related to loop teardown and orphaned tasks, because the context manager takes care of this when it exits. aioloop-proxy also includes an `advance_time` method, which has been proposed in #113.

I played around with aioloop-proxy a bit. A large number of tests already work when replacing a couple of lines in `wrap_in_sync`:
```
def wrap_in_sync(func, _loop):
…
        if coro is not None:
            try:
                with aioloop_proxy.proxy(_loop) as loop_proxy:
                    task = asyncio.ensure_future(coro, loop=loop_proxy)
                    loop_proxy.run_until_complete(task)
            except BaseException:
…
```

The failing tests were mostly related to subprocesses or setting a custom event loop or policy. Some special usage like the one in [TestEventLoopStartedBeforeFixtures](https://github.com/pytest-dev/pytest-asyncio/blob/master/tests/test_simple.py#L118-L137) break, because the tasks in the test cases get attached to a different loop.

Anyway, the new information made me wonder if the `event_loop` fixture still makes sense when using aioloop-proxy. 

That's all I got for now :)

asvetlov:
From my point of view, the plugin should support two modes.
1. In *compatible* mode the `event_loop` fixture works as-is; it is crucial for the sake of backward compatibility.
2. The *proxy* mode has no `event_loop` but works differently.

An overridable session-scope fixture exists that returns a *root loop instance*. An user can override this fixture in `conftest.py` to start a non-standard event loop, e.g. [`uvloop`](https://github.com/MagicStack/uvloop) or [`qasync`](https://github.com/CabbageDevelopment/qasync). The root fixture maker should be named differently than `event_loop` to don't clash with *compatible* mode.

[...]

PR #309, opened 2022-03-12
Add cleanup code to event_loop that mimics the behavior of asyncio.run

### LINKED ISSUE #200: Should the default event_loop fixture add task cleanup?
I'm not sure if this would be considered a bug or a feature :)

I noticed that in some of my tests, my long running tasks (e.g. while loops) were throwing warnings about tasks being destroyed but were pending. I guess this is happening because the loop is closed at the end of a test, but my while loops are still while'ing

In my projects, I maintain a cleanup function at the end of my main function to keep track of what tasks are still alive, then I cancel them before shutting down the loop. The tasks themselves are running forever intentionally, and they should only shut down when the app shuts down (and I shut down gracefully for cleaner logs). 

I was wondering if this functionality should be added to the event_loop fixture (maybe optionally) to silence warnings where they don't really belong? Or, is this the responsibility of the developer via a custom fixture?

Askaholic:
This is something I found useful and implemented in one of my own projects so I thought I'd make a PR in case the maintainers think this would be a good addition to the library itself. I'd be happy to modify it if only a subset of the changes is desired.

Some issues that I ran into with the current implementation:
- Errors that occur in a task spawned during one test can leak into subsequent tests since simply calling `loop.close()` is not sufficient to shutdown the event loop. This can make it extremely difficult to track down which test is causing the errors, since the test that fails is not actually the one that has the problem.
- When running with `-W error` under python 3.9 a similar problem happens with ResourceWarnings, where it becomes extremely difficult to figure out which test is actually opening the resources that are generating the warnings since the GC doesn't actually clean them up until some later test is already running.

Closes #200

### DIFF
--- a/pytest_asyncio/plugin.py
+++ b/pytest_asyncio/plugin.py
@@ -3,6 +3,7 @@ import asyncio
 import contextlib
 import enum
 import functools
+import gc
 import inspect
 import socket
 import sys
@@ -488,7 +489,19 @@ def event_loop(request: "pytest.FixtureRequest") -> Iterator[asyncio.AbstractEve
     """Create an instance of the default event loop for each test case."""
     loop = asyncio.get_event_loop_policy().new_event_loop()
     yield loop
-    loop.close()
+    # Cleanup code copied from the implementation of asyncio.run()
+    try:
+        asyncio.runners._cancel_all_tasks(loop)
+        loop.run_until_complete(loop.shutdown_asyncgens())
+        if sys.version_info >= (3, 9):
+            loop.run_until_complete(loop.shutdown_default_executor())
+    finally:
+        loop.close()
+        # Call the garbage collector to trigger ResourceWarning's as soon
+        # as possible (these are triggered in various __del__ methods).
+        # Without this, resources opened in one test can fail other tests
+        # when the warning is generated.
+        gc.collect()
 
 
 def _unused_port(socket_type: int) -> int:
--- a/pytest_asyncio/plugin.py
+++ b/pytest_asyncio/plugin.py
@@ -361,8 +361,15 @@ def pytest_fixture_post_finalizer(fixturedef: FixtureDef, request: SubRequest) -
         except RuntimeError:
             loop = None
         if loop is not None:
-            # Clean up existing loop to avoid ResourceWarnings
-            loop.close()
+            # Cleanup code based on the implementation of asyncio.run()
+            try:
+                if not loop.is_closed():
+                    asyncio.runners._cancel_all_tasks(loop)
+                    loop.run_until_complete(loop.shutdown_asyncgens())
+                    if sys.version_info >= (3, 9):
+                        loop.run_until_complete(loop.shutdown_default_executor())
+            finally:
+                loop.close()
         new_loop = policy.new_event_loop()  # Replace existing event loop
         # Ensure subsequent calls to get_event_loop() succeed
         policy.set_event_loop(new_loop)
@@ -487,22 +494,13 @@ def pytest_runtest_setup(item: pytest.Item) -> None:
 @pytest.fixture
 def event_loop(request: "pytest.FixtureRequest") -> Iterator[asyncio.AbstractEventLoop]:
     """Create an instance of the default event loop for each test case."""
-    loop = asyncio.get_event_loop_policy().new_event_loop()
-    yield loop
-    # Cleanup code copied from the implementation of asyncio.run()
-    try:
-        if not loop.is_closed():
-            asyncio.runners._cancel_all_tasks(loop)
-            loop.run_until_complete(loop.shutdown_asyncgens())
-            if sys.version_info >= (3, 9):
-                loop.run_until_complete(loop.shutdown_default_executor())
-    finally:
-        loop.close()
-        # Call the garbage collector to trigger ResourceWarning's as soon
-        # as possible (these are triggered in various __del__ methods).
-        # Without this, resources opened in one test can fail other tests
-        # when the warning is generated.
-        gc.collect()
+    return asyncio.get_event_loop_policy().new_event_loop()
+    # Call the garbage collector to trigger ResourceWarning's as soon
+    # as possible (these are triggered in various __del__ methods).
+    # Without this, resources opened in one test can fail other tests
+    # when the warning is generated.
+    gc.collect()
+    # Event loop cleanup handled by pytest_fixture_post_finalizer
 
 
 def _unused_port(socket_type: int) -> int:
--- a/pytest_asyncio/plugin.py
+++ b/pytest_asyncio/plugin.py
@@ -494,7 +494,7 @@ def pytest_runtest_setup(item: pytest.Item) -> None:
 @pytest.fixture
 def event_loop(request: "pytest.FixtureRequest") -> Iterator[asyncio.AbstractEventLoop]:
     """Create an instance of the default event loop for each test case."""
-    return asyncio.get_event_loop_policy().new_event_loop()
+    yield asyncio.get_event_loop_policy().new_event_loop()
     # Call the garbage collector to trigger ResourceWarning's as soon
     # as possible (these are triggered in various __del__ methods).
     # Without this, resources opened in one test can fail other tests
--- a/pytest_asyncio/plugin.py
+++ b/pytest_asyncio/plugin.py
@@ -491,10 +491,11 @@ def event_loop(request: "pytest.FixtureRequest") -> Iterator[asyncio.AbstractEve
     yield loop
     # Cleanup code copied from the implementation of asyncio.run()
     try:
-        asyncio.runners._cancel_all_tasks(loop)
-        loop.run_until_complete(loop.shutdown_asyncgens())
-        if sys.version_info >= (3, 9):
-            loop.run_until_complete(loop.shutdown_default_executor())
+        if not loop.is_closed():
+            asyncio.runners._cancel_all_tasks(loop)
+            loop.run_until_complete(loop.shutdown_asyncgens())
+            if sys.version_info >= (3, 9):
+                loop.run_until_complete(loop.shutdown_default_executor())
     finally:
         loop.close()
         # Call the garbage collector to trigger ResourceWarning's as soon
--- a/pytest_asyncio/plugin.py
+++ b/pytest_asyncio/plugin.py
@@ -364,7 +364,9 @@ def pytest_fixture_post_finalizer(fixturedef: FixtureDef, request: SubRequest) -
             # Cleanup code based on the implementation of asyncio.run()
             try:
                 if not loop.is_closed():
-                    asyncio.runners._cancel_all_tasks(loop)
+                    asyncio.runners._cancel_all_tasks(  # type: ignore[attr-defined]
+                        loop
+                    )
                     loop.run_until_complete(loop.shutdown_asyncgens())
                     if sys.version_info >= (3, 9):
                         loop.run_until_complete(loop.shutdown_default_executor())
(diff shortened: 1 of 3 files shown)


pytest_asyncio/plugin.py
```diff
@@ -484,7 +485,19 @@ def event_loop(request: "pytest.FixtureRequest") -> Iterator[asyncio.AbstractEve
     """Create an instance of the default event loop for each test case."""
     loop = asyncio.get_event_loop_policy().new_event_loop()
     yield loop
-    loop.close()
+    # Cleanup code copied from the implementation of asyncio.run()
+    try:
+        asyncio.runners._cancel_all_tasks(loop)
```
Askaholic, line 490: Calling private functions here is a bit sketch, but it's what I chose to do in my project instead of copy/pasting the entire implementation. Might want to go with the copy/paste approach though for this library.

seifertm:
Thanks for the contribution!

The linked issue is not the only request of that feature. There is also #222. We closed #222 in favour of #235, which intends to used `aioloop-proxy` to have a "virtualized" event loop.

The aioloop-proxy issue has not been started, as far as I know. I think it's a good idea to add clean up code for outstanding tasks independently of #235.

Could you add some tests that assert the correct behaviour of your code?

Askaholic:
Great! I've read through the issues you linked and I think I understand the current state of this feature. Sounds like eventually a more sophisticated solution will be added, but for now this PR will be useful. I'll get the development environment set up and work on writing a few tests!

asvetlov:
Regarding https://github.com/pytest-dev/pytest-asyncio/issues/235 (aioloop-proxy based solution).
I have a work-in-progress draft: https://github.com/pytest-dev/pytest-asyncio/pull/312
It is not finished yet, sorry. Now I'm working on improving asyncio itself, Python 3.11 feature freeze is coming.

pytest_asyncio/plugin.py
seifertm: Sorry for the late response… There are a couple of ideas flying around, all of which address this issue in some way. That makes it hard to review your PR, so I put it off for too long.

I agree that calling `_cancel_all_tasks` might not be a good idea. Although we test with different CPython versions, `_cancel_all_tasks` is an implementation detail that could change in a patch release. Since we only test different minor releases (i.e. 3.7, 3.8, …) a change of `_cancel_all_tasks` could slip through unnoticed.

It would be best to replicate the behaviour of `_cancel_all_tasks` in pytest-asyncio. We cant't do a 1:1 copy and paste, though, because the CPython code is protected by copyright and under a different license than pytest-asyncio.

pytest_asyncio/plugin.py
```diff
@@ -484,7 +485,20 @@ def event_loop(request: "pytest.FixtureRequest") -> Iterator[asyncio.AbstractEve
     """Create an instance of the default event loop for each test case."""
     loop = asyncio.get_event_loop_policy().new_event_loop()
     yield loop
-    loop.close()
+    # Cleanup code copied from the implementation of asyncio.run()
```
seifertm, line 488: This new code works great, thank you! Users that don't define their own `event_loop` fixture will benefit immediately from the cleanup code.

Users that do override `event_loop` would have to add a lot of code to their fixture definition, though.

Would it be possible to perform the cleanup in `pytest_fixture_post_finalizer`, fixture? That way, users who override `event_loop` would only have to remove  the line `loop.close()` to benefit from the task and generator cleanup. It would be a much smoother transition and less error prone.

tests/test_event_loop_cleanup.py
```diff
@@ -0,0 +1,38 @@
+from textwrap import dedent
+
+
+def test_task_canceled_on_test_end(testdir):
```
seifertm, line 4: Great job!

At first the test didn't seem to fail when I removed the new cleanup code. Then I noticed that we're testing with pytest-6.2.5. When I bumped the pytest version the test failed correctly.

Could you replace `pytest == 6.2.5` with `pytest >= 6.2.5` in the `[testenv]` of `tox.ini` while you're at it?

Askaholic: Hmm, that is strange. I'm running 7.1.1 locally.

Do I need to bump it in `[testenv:pytest-min]` too?

pytest_asyncio/plugin.py
Askaholic: Ok, I'll give it a go!

Askaholic: I tried this and it seemed to work except that I got some warnings in two of the tox environments:

```
tests/test_simple.py ..Exception ignored in: <function BaseEventLoop.__del__ at 0x7f2fe9fc5550>
Traceback (most recent call last):
  File "/usr/local/lib/python3.9/asyncio/base_events.py", line 681, in __del__
    _warn(f"unclosed event loop {self!r}", ResourceWarning, source=self)
ResourceWarning: unclosed event loop <TestEventLoop running=False closed=False debug=False>
Exception ignored in: <socket.socket fd=18, family=AddressFamily.AF_UNIX, type=SocketKind.SOCK_STREAM, proto=0>
Traceback (most recent call last):
  File "/home/user/Documents/pytest-asyncio/.tox/pytest-min/lib/python3.9/site-packages/pluggy/callers.py", line 169, in _multicall
    for hook_impl in reversed(hook_impls):
ResourceWarning: unclosed <socket.socket fd=18, family=AddressFamily.AF_UNIX, type=SocketKind.SOCK_STREAM, proto=0>
Exception ignored in: <socket.socket fd=19, family=AddressFamily.AF_UNIX, type=SocketKind.SOCK_STREAM, proto=0>
Traceback (most recent call last):
  File "/home/user/Documents/pytest-asyncio/.tox/pytest-min/lib/python3.9/site-packages/pluggy/callers.py", line 169, in _multicall
    for hook_impl in reversed(hook_impls):
ResourceWarning: unclosed <socket.socket fd=19, family=AddressFamily.AF_UNIX, type=SocketKind.SOCK_STREAM, proto=0>
```

These are the pytest sessions for which I got this warning:
```
platform linux -- Python 3.9.9, pytest-6.1.0, py-1.11.0, pluggy-0.13.1
cachedir: .tox/pytest-min/.pytest_cache
rootdir: /home/user/Documents/pytest-asyncio, configfile: setup.cfg, testpaths: tests
plugins: asyncio-0.18.3.dev7+gdddf3e4.d20220326, flaky-3.7.0, hypothesis-6.39.4
asyncio: mode=auto
collected 77 items
```

```
platform linux -- Python 3.10.0, pytest-7.1.1, pluggy-1.0.0
cachedir: .tox/py310/.pytest_cache
rootdir: /home/user/Documents/pytest-asyncio, configfile: setup.cfg, testpaths: tests
plugins: asyncio-0.18.3.dev7+gdddf3e4.d20220326, flaky-3.7.0, hypothesis-6.39.4
asyncio: mode=auto
collected 77 items
```

Both times it happened in the same place, after `test_asyncio_marker`. I also noticed that it's using an event loop called `TestEventLoop` which is set up in `tests/respect_event_loop_policy/conftest.py`, so that is possibly leaking out of the tests that it's actually intended for. Not sure if that's a problem.

tests/test_event_loop_cleanup.py
seifertm: The `pytest-min` environment was added to test pytest-asyncio against the lowest supported version of pytest. We shouldn't bump it without a proper deprecation warning well ahead of time.

pytest_asyncio/plugin.py
seifertm: > I tried this and it seemed to work except that I got some warnings in two of the tox environments

That sounds promising!

The `TestEventLoop` instance should not leak out of the `tests/respect_event_loop_policy` package.

It's hard for me to comment on the other two warnings, without getting my hands on the exact code you were running. Could you push the changes?

seifertm:
Thanks for uploading the latest version. I had some time to look into the issue that causes `ResourceWarning`s in the tests.

The culprit seems to be `test_event_loop_scope.py::test_4`. The test case requests the `event_loop` fixture and sets the current global loop to `None`. Requesting `event_loop` initializes a new loop. The code in `pytest_fixture_post_finalizer` tries to retrieve the current loop via `asyncio.get_event_loop_policy().get_event_loop()`, but that reference was set to `None` in the test case. The result is a stale event loop that triggers a `ResourceWarning` about an unclosed socket.

I was trying to work around the issue by using `fixturedef.cached_result` in `pytest_fixture_post_finalizer` to retrieve the loop provided by the `event_loop` fixture. Until now that wasn't successful, though. I'll keep digging…

pytest_asyncio/plugin.py
seifertm: The `ResourceWarning` was caused by `test_event_loop_scope.py::test_4` which sets an event loop to None without closing it. This seems to cause side-effects in other tests. I modified the test to close the loop before setting the current loop to None.

We should consider performing such tests with `pytester.runpytest_subprocess` to avoid side-effects on the rest of the test suite.

seifertm:
I modified `test_event_loop_scope.py::test_4` accordingly to avoid side-effects in other tests. The PR has been rebased and two typing issues have been addressed.

pytest_asyncio/plugin.py
seifertm: I still stand by my last comment on the use of `_cancel_all_tasks`.

pytest_asyncio/plugin.py
```diff
@@ -486,9 +496,13 @@ def pytest_runtest_setup(item: pytest.Item) -> None:
...
     """Create an instance of the default event loop for each test case."""
-    loop = asyncio.get_event_loop_policy().new_event_loop()
-    yield loop
-    loop.close()
+    yield asyncio.get_event_loop_policy().new_event_loop()
+    # Call the garbage collector to trigger ResourceWarning's as soon
+    # as possible (these are triggered in various __del__ methods).
+    # Without this, resources opened in one test can fail other tests
+    # when the warning is generated.
+    gc.collect()
```
seifertm, line 504: A bunch of people implement their own `event_loop` fixture, in order to control fixture scope. If we change the fixture definition, we'd have to inform them to adjust as well.

Asking users to remove `loop.close()` to benefit from the new cleanup code is an easy sell. Adding a call to `gc.collect` in the event_loop fixture not so much.

Is there away around the `gc.collect` call in the fixture code?

Askaholic: I suppose the call to `gc.collect` isn't directly related to task cleanup. I added this in my codebase because there were some tests that opened sockets and never closed them. Having the garbage collector run during the test teardown would make sure that the correct test failed when running pytest with `-W error`. But this is unrelated to pytest-asyncio so maybe it doesn't belong here.

elupus:
I've run into the issue of tasks leaking over test bounderings when i tried to get home assistant updated to later pytest-asyncio: So far i have these fixes/hacks: I've not managed to get it to run reliable as of yet thou.

Cleaning up of lingering tasks:
https://github.com/home-assistant/core/pull/82475/commits/6e9130dd90ce7c5ebdd580f047d29ae1636b81ee

Cleaning up of lingering timers:
https://github.com/home-assistant/core/pull/82475/commits/4d1335f75636475e45f423c1e3a0423230105528

elupus:
I also just noticed that aiohttp loop fixture did trigger a GC call after test finishes: https://github.com/aio-libs/aiohttp/blob/905e0625777c20e4cf2d57fcfd4355827312f797/aiohttp/test_utils.py#L525 

That likely explains some of my failures.

matzipan:
So are the open threads the only things left for this to go forward? Or is there something else missing? I'll try to find some time for this.

seifertm:
Closing this in favor of #1152 which leaves it to asyncio.Runner to clean up tasks when the underlying loop is out of scope.

PR #312, opened 2022-03-18
aioloop-proxy usage draft

asvetlov:


seifertm:
I'd be extremely happy if someone picked up this draft. I'd prefer to see three separate PR's though:
1. Introduce the concept of a "runner". Python 3.11 should use asyncio.Runner, whereas older Python version should use a backport of this.
2. Use of aioloop-proxy to allow nested loops (currently I'm a bit wary of integrating aioloop-proxy, as nobody besides Andrew seems to have access to the repo and to pypi and he's currently unavailable)
3. Adding timeout functionality for tests

The order of 2 and 3 are interchangeable :)

PR #404, opened 2022-09-01
Perform `in` (rather than `is) check against cached fixturedefs

DanCardin:
If i'm off base here, feel free to close this...

but in my struggles to get my plugin working with pytest-asyncio, i kept noticing this holder check; and i couldn't possibly fathom how a ` fixturedef` would come to **be** this `_HOLDER` dict.

but it would make a lot more sense if the intent here was to avoid reexecuting the logic further down, if you already have classified that fixturedef previously.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -2,6 +2,10 @@
 Changelog
 =========
 
+UNRELEASED
+=================
+- Fixed an issue which prevented fixture setup from being cached. `#404 <https://github.com/pytest-dev/pytest-asyncio/pull/404>`_
+
 0.19.0 (22-07-13)
 =================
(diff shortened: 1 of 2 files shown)


seifertm:
> If i'm off base here, feel free to close this...
Not at all. Your patch is much appreciated!

The intent is indeed to avoid executing the remainder of `_preprocess_async_fixtures`. I believe the `is` check is unintentional. It is also my understanding that this if clause can never be `True`.

Good find!

seifertm (changes requested):
Could you add an entry to the changelog and link the PR? I think it's worthwhile to mention that you fixed an issue that prevented the fixture setup to be cached.

seifertm (approved):
Thank you!

PR #414, opened 2022-10-02
Remove legacy mode

seifertm:
Asyncio modes have been introduced in v0.17.0 which was released on 2022-01-13. In addition to the auto and strict modes, the release defaulted to "legacy" mode for backwards compatibility. The release also issued a deprecation warning when using legacy mode.

The default asyncio mode was changed to _strict_ in the v0.19.0 release on 2022-07-13.

This change removes legacy mode entirely. Users of pytest-asyncio have had a 9 month deprecation cycle at the time of writing.


Tinche (approved):
Lgtm

PR #432, opened 2022-10-21
Fix setuptools deprecation warning for license_file

mgorny:
Use `license_files` instead of `license_file` in `setup.py` to fix the following warning:

> The license_file parameter is deprecated, use license_files instead.


seifertm:
Thanks, that deprecation has hasn't been noticed, yet.

The new option is supported since setuptools v42 and we require >=51 so this shouldn't be an issue.

PR #439, opened 2022-11-06
Handle bound fixture methods correctly

### LINKED ISSUE #197: async method fixture is run with different `self` than the test method
Hi,
I just ran into this - if you make a fixture method `async`, it is run with a different `self`, so you can't set attributes on `self` and read them in test methods. Here is a simple snippet to reproduce the problem:

```python
import pytest

pytestmark = pytest.mark.asyncio

class TestAsyncFixture:
  @pytest.fixture #(autouse = True)
  async def setup(self):
    # await self.client._handle_ws_request(self.payment_start_request)
    self.some_attr = 'value'

    print()
    print('in fixture')
    print(self)
    print(self.__dict__)

  async def test_response(self, setup):
    print('in test')
    print(self)          # different object
    print(self.__dict__) # some_attr missing
```

At first, I thought this only concerns `autouse` fixtures, but now I tried using it without `autouse` and it has the same problem. Is this a known issue, or is it supposed to work?

I'm using
```
python            3.8.3
pytest            6.1.2
pytest-asyncio    0.14.0
```

Btw, thanks for this great library, I'm using it heavily and it's been of great help.

mjpieters:
When passed a _bound_ fixture method, re-bind that method to the currently active instance.

The implementation follows how [pytest handles bound fixtures](https://github.com/pytest-dev/pytest/blob/2f33ea87c8d303c3439655eb2bea24483bffde5e/src/_pytest/fixtures.py#L1086-L1110), including how it'll always bind any fixture when `fixturedef.unittest` is true, and otherwise only if the fixture wasn't already bound or if bound to a an instance of the same class or a subclass thereof.

Fixes #197

### DIFF
--- a/pytest_asyncio/plugin.py
+++ b/pytest_asyncio/plugin.py
@@ -227,11 +227,10 @@ def _synchronize_async_fixture(fixturedef: FixtureDef) -> None:
     """
     Wraps the fixture function of an async fixture in a synchronous function.
     """
-    func = fixturedef.func
-    if inspect.isasyncgenfunction(func):
-        fixturedef.func = _wrap_asyncgen(func)
-    elif inspect.iscoroutinefunction(func):
-        fixturedef.func = _wrap_async(func)
+    if inspect.isasyncgenfunction(fixturedef.func):
+        _wrap_asyncgen_fixture(fixturedef)
+    elif inspect.iscoroutinefunction(fixturedef.func):
+        _wrap_async_fixture(fixturedef)
 
 
 def _add_kwargs(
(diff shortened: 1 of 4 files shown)


seifertm:
The CI runs the tests with `-W error`. For some reason, it pulled in the latest version of pytest. v7.2 introduces some deprecation warnings which, in turn, triggered errors. Therefore, I had to merge #438 first and rebase this PR.

pytest_asyncio/plugin.py
```diff
@@ -249,14 +248,38 @@ def _add_kwargs(
     return ret
 
 
-def _wrap_asyncgen(func: Callable[..., AsyncIterator[_R]]) -> Callable[..., _R]:
-    @functools.wraps(func)
+def _perhaps_rebind_fixture_func(
+    func: _T, instance: Optional[Any], unittest: bool
```
seifertm, line 252: The type checker complains that `_T` doesn't have the `__func__`, `__self__`, and `__get__` attributes. We need to address this before merging the PR.

I'm not aware of a proper type bound for `_T` that would fix the issue (or even a type that expresses a bound method). Ideas are welcome, but IMHO it would be acceptable to ignore the warnings inside of `_perhaps_rebind_fixture_func`.

pytest_asyncio/plugin.py
```diff
@@ -249,14 +248,38 @@ def _add_kwargs(
     return ret
 
 
-def _wrap_asyncgen(func: Callable[..., AsyncIterator[_R]]) -> Callable[..., _R]:
-    @functools.wraps(func)
+def _perhaps_rebind_fixture_func(
```
seifertm, line 251: You described two ways to fix the issue in #197: Adjusting `_wrap_async` so that it looks like a bound method to pytest or reimplement pytest's method rebinding.

Can you share the reasoning why you went with reimplementing the method rebinding rather than adding `__self__` to `_wrap_async_`?

seifertm (changes requested):
This is a great PR. Thank you!

pytest_asyncio/plugin.py
mjpieters: This was the more elegant path, I felt. Not a very satisfying answer, as it is subjective. Mostly, I wasn't sure I understood all the edgecases of the interaction between pytest, the wrapper, and both bound and unbound fixture functions. By using this approach you can follow how the fixture function is rebound or not entirely within the pytest-asyncio codebase, so the process is easier to follow and reason about.

seifertm: That is a perfectly satisfying answer to me, thank you.

PR #452, opened 2022-11-22
Fix usage of anyio TaskGroups in fixtures

nikicat:
When I use async generator fixutres with 
```
async with anyio.create_task_group() as tg:
    ...
``` 
inside I get this error on exit: `Attempted to exit cancel scope in a different task than it was entered in`
This patch makes a single task for the fixture enter and exit.
May be there is a more simple and elegant way to solve this (without using `Queue`) but this is what I found.

### DIFF
--- a/pytest_asyncio/plugin.py
+++ b/pytest_asyncio/plugin.py
@@ -283,22 +283,53 @@ def _wrap_asyncgen_fixture(fixturedef: FixtureDef) -> None:
             res = await gen_obj.__anext__()
             return res
 
+        async def async_finalizer() -> None:
+            try:
+                await gen_obj.__anext__()
+            except StopAsyncIteration:
+                pass
+            else:
+                msg = "Async generator fixture didn't stop."
+                msg += "Yield only once."
+                raise ValueError(msg)
+
+        in_event = asyncio.Event()
+        out_queue = asyncio.Queue()
+
+        async def fixture_runner():
+            try:
+                result = await setup()
+            except Exception as exc:
+                out_queue.put_nowait((None, exc))
+            else:
+                out_queue.put_nowait((result, None))
+            await in_event.wait()
+            try:
+                await async_finalizer()
+            except Exception as exc:
+                out_queue.put_nowait(exc)
+            else:
+                out_queue.put_nowait(None)
+
+        task = event_loop.create_task(fixture_runner())
+
         def finalizer() -> None:
             """Yield again, to finalize."""
 
-            async def async_finalizer() -> None:
+            async def finalize():
+                in_event.set()
                 try:
-                    await gen_obj.__anext__()
-                except StopAsyncIteration:
-                    pass
-                else:
-                    msg = "Async generator fixture didn't stop."
-                    msg += "Yield only once."
-                    raise ValueError(msg)
+                    exc = await out_queue.get()
+                    if exc is not None:
+                        raise exc
+                finally:
+                    await task
 
-            event_loop.run_until_complete(async_finalizer())
+            event_loop.run_until_complete(finalize())
 
-        result = event_loop.run_until_complete(setup())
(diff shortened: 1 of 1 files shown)


seifertm:
@nikicat Thanks for filing a PR and sorry for the long wait.

Frankly, the reason why it took so long to answer is because it's hard for me to judge the solution when I don't yet understand the underlying problem. Before I know what's wrong, I cannot really merge or close the PR.

In order to clear this out, I'd ask you open an issue that describes the problem you're solving with this PR. The issue should also include a [minimal example](https://stackoverflow.com/help/minimal-reproducible-example) that reproduces the issue. This allows me to investigate and understand the problem in depth.

I really appreciate you providing a fix, but I hope you also understand my problem in handling your PR. In general, filing an issue before a PR is the best approach, unless you're making trivial changes.

nikicat:
[Here it is](https://github.com/nikicat/pytest-asyncio-452/blob/master/tests/test_one.py)

seifertm:
Thanks for providing an example for me to play around. One thing I'd like to note is that you configured pytest-asyncio to run in *auto* mode. This will always cause issues when your test suite uses any other async plugin (e.g. anyio or pytest-trio; see [Test discovery modes](https://pytest-asyncio.readthedocs.io/en/v0.20.3/concepts.html#test-discovery-modes)). For the remainder of this comment, I'll assume that pytest-asyncio runs in strict mode.

You're proposing that anyio task groups should be able to run in a pytest-asyncio managed test run. The following code block tries to illustrate the fixtures and tests and the managing library is printed in parentheses:
```
event_loop (pytest-asyncio)
    + agen_fixture (pytest-asyncio)
        + task group (anyio)
            + the_task (*)
    + test_aaa (pytest-asyncio)
```

However, I don't understand why this setup is necessary. Anyio provides a runtime environment and a corresponding pytest plugin that generalizes over different async frameworks. That means you should be able to run any asyncio code with anyio. There's no need for pytest-asyncio at all.

The following code is based on your linked example and runs as expected when using anyio tests. Pytest-asyncio is completely eliminated from the tests:
```
import asyncio
import anyio
import pytest

@pytest.fixture
def anyio_backend():
    return 'asyncio'

async def the_task():
    await asyncio.sleep(1)

@pytest.fixture
async def agen_fixture():
    async with anyio.create_task_group() as tg:
        tg.start_soon(the_task)
        yield

@pytest.mark.anyio
async def test_aaa(agen_fixture):
    pass
```

Given that anyio allows running asyncio code, I don't see a use case for mixing pytest-asyncio with anyio. The PR also introduces a significant amount of complexity distributed in different places of pytest-asyncio due to its use of *asyncio.Event* and *asyncio.Queue.* Unless there's a very compelling use case for using pytest-asyncio alongside anyio tests, this feature will not be integrated.

I'm sorry that you put in all that effort to address the issue and read this response. Feel free to continue the discussion and elaborate your use case, if you think I'm missing something.

nikicat:
Thank you for the detailed answer, @seifertm 
In my original case fixtures didn't create anyio task group directly, there was an intermediate layer, so it was not very convenient to mark tests with `@pytest.mark.anyio` depending on a fixture that invoke some other code that could potentially invoke `anyio.create_task_group()`.
I agree that my fix is pretty complicated and ugly and maybe it should better be solved in anyio than here.
Anyway, I've already migrated my project to Python 3.11 where I can use `asyncio.TaskGroup()` instead of anyio, so the problem will be solved.

didimelli:
Hey there @seifertm, sorry to jump back on this closed PR. I am trying the exact snippet from your comment above and I still get the same exact error `RuntimeError: Attempted to exit cancel scope in a different task than it was entered in` (I am using `anyio==4.3.0`). Do you have any idea why that would happen?

seifertm:
@didimelli The error message doesn't ring a bell. Do you mind creating a separate issue with a small reproducible example?

PR #470, opened 2022-12-22
python 3 changes

marksmayo:
- sorted and tidied imports
 - added trailing commas where 3.5, 3.6+ require
 - simplified some conditional statements

### DIFF
--- a/pytest_asyncio/plugin.py
+++ b/pytest_asyncio/plugin.py
@@ -7,22 +7,9 @@ import inspect
 import socket
 import sys
 import warnings
-from typing import (
-    Any,
-    AsyncIterator,
-    Awaitable,
-    Callable,
-    Dict,
-    Iterable,
-    Iterator,
-    List,
-    Optional,
-    Set,
-    TypeVar,
-    Union,
-    cast,
-    overload,
-)
+from typing import (Any, AsyncIterator, Awaitable, Callable, Dict, Iterable,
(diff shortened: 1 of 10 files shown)


Pierre-Sassoulas:
Isn't it something that could be automated and enforced with isort/pre-commit/black ?

seifertm:
@marksmayo Thanks for reaching out and submitting a PR.

> sorted and tidied imports

> Isn't it something that could be automated and enforced with isort/pre-commit/black ?

Indeed. Pytest-asyncio uses pre-commit to execute [shed](https://pypi.org/project/shed/), which in turn runs isort and black to organize imports and format the code base. This is also checked as part of the CI (see [the CI run for this PR](https://github.com/pytest-dev/pytest-asyncio/actions/runs/3754774589/jobs/6384998527)). There should be no need to perform manual formatting changes.

> added trailing commas where 3.5, 3.6+ require

Python 3.5 and Python 3.6 are end-of-life. Unless there's a very compelling reason, I don't think we'll accept pull requests that make changes for unmaintained Python versions.

> simplified some conditional statements

This would be actually helpful. I'm happy to accept the kinds of changes if it makes the code more readable and easier to understand.

@marksmayo I'll close this PR for the above mentioned reasons. However, I invite you to reopen it, if you want to simplify conditionals. Just keep in mind to install [pre-commit](https://pre-commit.com/) and rund `pre-commit install` before developing. This will automatically format all the code and sort imports according to the current coding standards of pytest-asyncio.

PR #480, opened 2023-01-15
Fail CI run when the corresponding Python environment cannot be found

seifertm:
This avoids successful CI runs for Python environment which never actually execute test code. One example where this happened in the past was addressed in commit 25d9592286682bc6dbfbf291028ff7a9594cf283.


seifertm:
Closing this PR, as the behavior that's supposed to be fixed could not be reproduced.

PR #496, opened 2023-02-13
Add documentation link after migration to docs

lbt:
The README doesn't provide an easy way to find the docs :)


seifertm (commented):
That makes sense. Thanks for correcting this oversight @lbt!

PR #588, opened 2023-07-23
Add how-to guide for testing with uvloop

seifertm:
This PR adds a how-to guide for testing with uvloop.

The guide is being added as a follow up to the discussion #587.


Tinche (approved):

seifertm:
Setting the loop policy inside a fixture may cause issues ( see #591). Therefore, I adjusted the how-to guide to set the policy directly in _conftest.py_ rather than in a pytest fixture.

Issue #591, opened 2023-07-25
Unexpected behavior when defining custom loop policy


schlamar:
When defining a custom loop policy in a fixture, the first test is executed with the loop from the default event loop policy and not from the custom loop policy. I assume pytest-asyncio creates the first event loop before the fixture for the custom loop policy has run.

Simple reproduction test (Windows):

```python
import asyncio
import pytest


@pytest.fixture(autouse=True, scope="module")
def set_loop_policy():
    asyncio.set_event_loop_policy(asyncio.WindowsSelectorEventLoopPolicy())


@pytest.mark.asyncio
async def test_loop(event_loop):
    assert isinstance(event_loop, asyncio.SelectorEventLoop)


@pytest.mark.asyncio
async def test_loop_second(event_loop):
    assert isinstance(event_loop, asyncio.SelectorEventLoop)
```

The first test fails, the second passes. I would expect that both tests pass.

Related to #587 and #588

seifertm:
Thanks for the report and the reproducer!

There seems to be a problem with the order in which fixtures are resolved. The _event_loop_ fixture is evaluated before  _set_loop_policy_, so the policy change only takes effect after the first loop has been created. You can check for yourself by adding the `--setup-show`` option to the pytest call.

As a workaround, you can call `asyncio.set_event_loop_policy` outside of the fixture (i.e. directly in the module).

As per the [pytest docs,](https://docs.pytest.org/en/stable/reference/fixtures.html#fixture-instantiation-order) wider-scoped fixtures are evaluated earlier. This behavior is probably broken by the [_event_loop_ fixture injection](https://github.com/pytest-dev/pytest-asyncio/blob/v0.21.1/pytest_asyncio/plugin.py#L551-L553) performed by pytest-asyncio.

seifertm:
Changing pytest-asyncio to evaluate the _event_loop_ fixture last rather than first would solve the issue. However, the _event_loop_ fixture technically is an autouse fixture, so it should be evaluated together with other autouse fixtures of the same scope.

seifertm:
My understanding is that only the first test uses the wrong event loop policy. If you really need the policy to be set in a pytest fixture, you can add a dummy test in between the policy fixture and the tests as a workaround:
```python
import asyncio
import pytest

class MyLoop(asyncio.SelectorEventLoop):
    pass

class MyLoopPolicy(asyncio.DefaultEventLoopPolicy):
    def new_event_loop(self):
[...]

PR #620, opened 2023-09-20
Scoped event loops based on pytest marks

seifertm:
This PR introduces the _asyncio_event_loop_ mark. When applied to classes and modules, the mark provides an asyncio event loop that is class-scoped or module-scoped, respectively. This is a prerequisite for the deprecation of event_loop fixture overrides (see #631).

Fixtures and tests under the same _asyncio_event_loop_ mark are run in the same event loop, thus addressing #127 in parts.

The _asyncio_event_loop_ mark supports an optional `policy` keyword argument, which takes an event loop policy or an iterable of policies. Tests under the effect of the mark are parametrized over the provided policies. If no policy is provided, the mark falls back to the value of `asyncio.get_event_loop_policy()` at mark definition time. This functionality offers a workaround for #591.

The implementation dynamically creates pytest fixtures based on the presence of the _asycnio_event_loop_ mark. The fixtures are anchored at the corresponding module or class. The generated fixtures have a unique name that is generated from the pytest node ID of the module or class.

The generated fixtures are always parametrized, even when just a single event loop policy is specified. Therefore, we need to make pytest aware which parameters to use when evaluating the fixture. This happens in `pytest_generate_tests`. Whenever _pytest_generate_tests_ is called for a test that is affected by an _asyncio_event_loop_ mark_, the name of the dynamic fixture is injected into the metafunc along with the dynamic fixture definition. This information allows pytest to perform the standard fixture parametrization.

There is one special case, where a user may have requested the `event_loop` fixture for an async test that also requests a scoped event loop. In this case, the code raises an error, because we cannot possibly know what the user intended to do here.


seifertm:
Added support for specifying a single event loop policy as a keyword argument to the `asyncio_event_loop` mark.

seifertm:
* Added support for multiple policies to be specified via the `asyncio_event_loop` mark. Tests under the same mark are parametrized accordingly.
* Deprecated overrides of the `event_loop` fixture in favour of the `asyncio_event_loop` mark.

seifertm:
I decided to move the deprecation of the event loop fixture into a separate PR.
This PR is a feature PR only.

PR #637, opened 2023-10-21
Group test logic with dedicated pytest.Item types

seifertm:
Pytest-asyncio implements different logic for running tests based on coroutines, static coroutines, and even async Hypothesis tests. For example, static coroutines need to be unwrapped before the test can be run. The logic for handling each case is distributed between the hooks `pytest_collection_modifyitems` and `pytest_pyfunc_call`. Both places require if-else clauses to decide which logic to apply for the current test item. This not only leads to duplicated code, but also makes the logic more difficult to follow, as it is distributed in several places of the plugin.

The goal of this PR is to concentrate the logic for each type of test into a single place.

Pytest provides the `pytest.Function` item that implements `setup` and `runtest`. These functions are invoked in the respective test run phases. This PR introduces `PytestAsyncioFunction`, a subclass of `pytest.Function`, which is the base class for all test items managed by pytest-asyncio. The new implementation contains one subclass of `PytestAsyncioFunction` for each type of test item that needs special treatment (i.e. async Hypothesis tests, async staticmethods, and coroutines). The logic for handling each of the cases is therefore limited to the scope of the respective subclass.

### DIFF
--- a/dependencies/default/constraints.txt
+++ b/dependencies/default/constraints.txt
@@ -20,4 +20,5 @@ sortedcontainers==2.4.0
 tomli==2.0.1
 trio==0.22.2
 typed-ast==1.5.5
+typing-extensions==4.7.1
 zipp==3.17.0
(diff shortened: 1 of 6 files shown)


seifertm:
Removed use of `typing.Self` in internal method, which avoids the dependency on _typing_extensions_.

Issue #657, opened 2023-10-31
Other event loop scopes?


2e0byo:
The recent changes deprecated redefining the event_loop fixture, but provided no replacement for `scope=package`.  Presumably the move away from the fixture was taken for good reasons, but I'd like to suggest support for (at least) package scope is worth having.  

Package scope is really handy: you can have a directory structure like this:
```
a/
├─ test_foo.py
├─ test_bar.py
b/
├─ test_baz.py
├─ test_blah.py
```
with expensive fixtures per-directory, spin them up once per dir and then clean them up (freeing ram) before moving to the next dir.

I know this will work just fine with a session scoped event loop (which is also no longer possible?), but matching the event loop scope to the context has advantages:

- every new testing context (new package) starts in a clean loop
- failing fixtures or leftover state (locks, unawaited coros) throw at the package level, so you get debugging quickly
- certain exotic failure modes where stale state on the loop is inadvertently cleaned up elsewhere (ahem) *only* show if the loop is explicitly torn down on exiting the context, effectively asserting nothing is running on it

Thus I'd like to suggest support for package scope is added to `asyncio_event_loop`.  For completeness, the  scopes are:

- function: default
- class: already available
- module: already available
- package: missing
- session: missing

I also think it is worth implementing session scoping, although I think `package` is normally the right scope.  It's possible I'm missing something here and this is already possible.

seifertm:
That's a fair point.

I deliberately omitted session scope, but I wasn't aware of package scope.

Do you have suggestion how package-scoped (and session-scoped) loops are defined?

val-melev:
what's the rationale for omitting session scope?

seifertm:
> what's the rationale for omitting session scope?

There were essentially two ideas for replacing event loop fixture overrides. The first was to have a config option (similar to `asyncio_mode`) that defines the scope of the event loop. However, this would force all tests in a project to use the same event loop scope. Since a single project can have a mix of small "unit" tests and larger tests (e.g. involving starting a web server), this approach seemed too limiting.

The other option was to tie the loop scope to a pytest mark. I'm not aware of a way for users to add a mark to the whole session, unless they know about pytest hooks. So in a way, it was an accepted drawback of the current solution.

@val-melev Do you have a use case that requires session-scoped loops? If so, can you share it?

2e0byo:
> I deliberately omitted session scope, but I wasn't aware of package scope.

[...]

PR #662, opened 2023-11-03
Introduce event_loop_policy fixture

### LINKED ISSUE #591: Unexpected behavior when defining custom loop policy

seifertm:
This PR introduces the new fixture _event_loop_policy._ The fixture returns an instance of an _asyncio.AbstractEventLoopPolicy_. By default, the fixture has _session_ (i.e. it is accessible by all tests) and returns the value of _asyncio.get_event_loop_policy()_.

Users can run pytest-asyncio tests with different types of event loops by redefining the fixture to return a different event loop policy. Parametrizing the fixture causes pytest-asyncio tests to be run once for each specified policy. This allows writing tests against multiple different types of event loops.

Although the concept of event loop policies is likely going to be [deprecated in CPython 3.13,](https://github.com/python/cpython/issues/94597) it was decided that this functionality is still going to land in pytest-asyncio. Starting with CPython 3.13, pytest-asyncio will provide a migration path for the deprecated policy system.

Resolves #591 
Prerequisite for #657

### DIFF
--- a/docs/source/how-to-guides/index.rst
+++ b/docs/source/how-to-guides/index.rst
@@ -5,6 +5,7 @@ How-To Guides
 .. toctree::
   :hidden:
 
+  multiple_loops
   uvloop
 
 This section of the documentation provides code snippets and recipes to accomplish specific tasks with pytest-asyncio.
--- a/docs/source/how-to-guides/multiple_loops.rst
+++ b/docs/source/how-to-guides/multiple_loops.rst
@@ -1,0 +1,10 @@
+======================================
+How to test with different event loops
+======================================
+
+Parametrizing the *event_loop_policy* fixture parametrizes all async tests. The following example causes all async tests to run multiple times, once for each event loop in the fixture parameters:
+
+.. include:: multiple_loops_example.py
+    :code: python
+
+You may choose to limit the scope of the fixture to *package,* *module,* or *class,* if you only want a subset of your tests to run with different event loops.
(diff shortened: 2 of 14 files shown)


Tinche:
Since I'm active on DPO, I just wanted to make sure you're aware of tentative plans to deprecate event loop policies in general (https://discuss.python.org/t/removing-the-asyncio-policy-system-asyncio-set-event-loop-policy-in-python-3-15/37553). From what I can tell, nothing has been decided yet, just bringing this to your attention.

Although I think the idea here is sound. If event loop policies are deprecated, we could allow users to override a different fixture (`event_loop_configuration`?) to apply simple customizations to event loops without overriding the `event_loop` fixture itself.

seifertm:
Thanks for mentioning the policy deprecation!

The topic was also touched in https://github.com/pytest-dev/pytest-asyncio/issues/657#issuecomment-1792465874 ff.
My takeaway from this thread was that we don't need to deprecate things faster than upstream.

If you also think the approach is generally sound, I'll continue to pursue this.

PR #696, opened 2023-11-30
Remove test dependencies: flaky, mypy

### LINKED ISSUE #674: Remove dependency on mypy
Mypy is listed as a dependency of pytest-asyncio under the `testing` extra. Mypy is not a runtime dependency of pytest-asyncio, nor is it required to run the tests.

Mypy should be removed from the testing dependencies and the `requirements.txt` and `constraints.txt` files should be adjusted accordingly.

### LINKED ISSUE #678: Remove dependency on flaky
Flaky is currently a dependency in the `testing` extra of pytest-asyncio. There are dedicated tests to test the compatibility between pytest-asyncio and flaky.

Pytest-asyncio should be compatible with other plugins, but it should not be responsible to test that compatibility. If every pytest plugin wrote tests against other plugins, we ended up in dependency hell and a lot of work would be done multiple times.

There's a dedicated section in pytest for testing plugin integration. This is the right place to do it.

Moreover, the flaky project seems to be abandoned. https://github.com/box/flaky/issues/192

The goal of this issue is to remove flaky from the the project dependencies.

touilleWoman:
I removed flaky, mypy.
I removed also mypy-extensions, typing_extensions which are dependencies of mypy.
Although tomli is dependency of mypy, I didn't remove it because Pytest depends on tomli.

Resolves #674
Resolves #678


seifertm:
Thanks for the PR!

I mentioned the removed dependencies in the changelog, because they're relevant for downstream packagers.

seifertm (approved):

Issue #708, opened 2023-12-06
Loop is closed before fixture teardown completes


jpwright:
I am looking to use pytest-asyncio to create a fixture that spawns a subprocess and then kills the subprocess when the fixture teardown is complete. I've noticed that the event loop can be closed before the fixture teardown is complete, which can lead to the subprocess kill not working.

Here is a simple example (tested using pytest-asyncio 0.23.2):

```python
import asyncio
import pytest
import pytest_asyncio

import logging
logger = logging.getLogger(__name__)

@pytest_asyncio.fixture
async def echo():
    p = await asyncio.create_subprocess_shell(
        f"sleep 2"
    )
    yield p
    p.kill()

@pytest.mark.asyncio
async def test_echo(echo):
    logger.info(echo)

```

Which produces the following output:

```bash
================================ test session starts =================================
platform linux -- Python 3.11.6, pytest-7.4.3, pluggy-1.3.0
rootdir: /home/jason/pytest-asyncio
plugins: asyncio-0.23.2
asyncio: mode=Mode.STRICT
collected 1 item                                                                     

test.py::test_echo 
----------------------------------- live log call ------------------------------------
INFO     test:test.py:18 <Process 756596>
PASSED                                                                         [100%]
--------------------------------- live log teardown ----------------------------------
WARNING  asyncio:unix_events.py:1427 Loop <_UnixSelectorEventLoop running=False closed=True debug=False> that handles pid 756596 is closed


================================= 1 passed in 0.00s ==================================
```

I am invoking the test using `python3 -m pytest --log-cli-level=INFO ./test.py`. I noticed that the issue is intermittent without logging to CLI... perhaps timing-related.

In my actual use case the warning cannot be ignored as the subprocess will not terminate on its own.

jpwright:
Thinking this over a bit more, I guess what I need is a way to insert my fixture's finalizer before the event_loop finalizer to the pytest request.

tonal:
also https://github.com/pytest-dev/pytest-asyncio/issues/706#issuecomment-1842106981

seifertm:
@jpwright Can you confirm that your issue is resolved with `pytest-asyncio==0.23.3a0`?

jpwright:
I do still get the `WARNING  asyncio:unix_events.py:1427 Loop <_UnixSelectorEventLoop running=False closed=True debug=False> that handles pid <...> is closed` message with `pytest-asyncio==0.23.3a0`.

[...]

PR #709, opened 2023-12-06
Bug fix: Issue #708

Franco-from-Owlish:
During teardown, the final test fails due to the event loop already being closed. 

The PR addresses that by checking if the look is open in the teardown function.


seifertm (changes requested):
Thanks for the initiative!
I believe this would fix the issue for many users, but there are some edge cases that are not addressed by the patch.

The underlying problem is not just that the fixture finalizer doesn't have a working event loop. It's that anything after he `yield` of an async generator fixture has no working event loop when using a (non-function-scoped) fixture with a function-scoped test.

The following example would still fails with the proposed patch, because `await task` raises an "attached to different loop" RuntimeError:
```python
import pytest
import pytest_asyncio

@pytest_asyncio.fixture(scope="module")
async def async_generator_fixture():
    task = asyncio.create_task(asyncio.sleep(0))
    yield
    await task


@pytest.mark.asyncio
async def test_asdf(async_generator_fixture):
    pass
````

The problematic piece of code lies in the setup of the `event_loop` fixture:
https://github.com/pytest-dev/pytest-asyncio/blob/b614c77dec6df7414fba78d2fbe7989c6ee16828/pytest_asyncio/plugin.py#L749-L753

We cannot just remove `old_loop.close()`, because that would break backwards compatibility. The v0.23 release is intended to introduce a bunch of deprecations to give users the chance to update their test suites accordingly.

I was thinking that we could attach a special attribute to the event loops provided by pytest-asyncio and use that attribute to determine if we should close the event loop or not. We can be sure that the loops provided by pytest-asyncio are cleaned up properly, so we skip the closing logic in this case, in order to prevent cleaning up the loop prematurely. A similar approach exists in the `event_loop` fixture implementation of pytest-asyncio:
https://github.com/pytest-dev/pytest-asyncio/blob/b614c77dec6df7414fba78d2fbe7989c6ee16828/pytest_asyncio/plugin.py#L951-L956

This is pretty hacky, but it's only a temporary solution for the v0.23 release.

And of course the PR would need to a test that reproduces the error along with a fix :)

seifertm:
Closing this in favor of #714 

I hope this doesn't discourage you from future contributions.

Issue #729, opened 2023-12-24
pytest-asyncio-0.23.2 loads packages more eagerly


seifertm:
In our case it is actually causing issues. In short, we have Python files in our tests dir that should only be loaded in a specific context. I'll elaborate in more detail below. We could probably work around it if we need to, but for now we have pinned pytest-asyncio to the previous version until this discussion reaches a conclusion.

Our specifix use case: we are developing a Python project with a plugin system. Put simply, these plugins are Python packages that register themselves at load time. They are expected to be loaded by a custom importlib finder/loader and they do not work as standalone package.

Since this plugin system is an integral part of the application we're developing, we test their behavior thoroughly. Therefore we have a bunch of these plugin "packages" as resources underneath our `tests` directory. They are meant as nothing more than data: our application will find and load them when appropriate. They will raise an error when loaded as a standalone package, which pytest v0.23 now does.

_Originally posted by @sanderr in https://github.com/pytest-dev/pytest-asyncio/issues/713#issuecomment-1868484144_

xmatthias:
We're seeing something similar (i suspect) - though due to an even odder case.

Our tests work fine up to pytest-asyncio-0.23.x ... at which point, ONLY the mac runs are failing (they do fail with `Fatal Python error: Segmentation fault`).
Based on my tests, this is related to the `torch` module (uninstalling torch completely will fix the issue).

Now `torch` (or pytorch) is not an unproblematic package on it's own.
it's doing some funky stuff when imported (it messes with logging, for one - which can be triggered by simply running `import torch`).

We do have an auto-use fixture to _completely_ mock the torch module on macos to work around this problem - but (i suspect) due to the eager loading of pytest-asyncio, it's loading torch BEFORE actually running the auto-use fixture, therefore importing torch - which causes the random segfaults.

I could also confirm this by messing with the installed torch file (simply placing a `raise ValueError()` in the `__init__.py` file confirmed that `pytest-asyncio==0.21.1` doesn't import this file, while `pytest-asyncio==0.23.2` does during the collection phase.

It's unclear why pytest-asyncio (a plugin to provide async tests, not to modify the test sequence, or help find additional tests) does need to mess with the importing logic and would import packages that

* don't have a direct import in any test file
* are not imported in a nested way

Now torch is imported in some files - which are only referenced through a custom resolver responsible to load the package (similar to the original author of this issue) - but these files are not touched or imported directly anywyere.


[...]

PR #747, opened 2024-01-14
No virtual modules & No scoped loops for unknown collectors.

seifertm:
This PR changes the way package-scoped fixtures are registered. Pytest-asyncio no longer creates temporary Python modules (aka "virtual modules") with fixture definitions during collection time. Some setups seem to rely on a fixed list of collected files (see https://github.com/pytest-dev/pytest-asyncio/issues/729#issuecomment-1887800329).

This PR also limits the impact of pytest-asyncio's collection changes for scoped event loops. The plugin no longer tries to install scoped event loops in unknown collector types.


seifertm:
Fixes #729

PR #769, opened 2024-01-30
Fix typing and update to mypy 1.8.0

bluetech:
When reading the code I noticed the line `_event_loop_fixture_id = StashKey[str]` which should be `_event_loop_fixture_id = StashKey[str]()` (an instance). I wondered how come mypy didn't complain, and the reason is that pytest wasn't installed in the pre-commit mypy environment.

This PR:
- Adds the missing pytest dep
- Fixes the resulting typing errors
- Upgrades mypy to the latest version
- Fixes `mypy tests/` by makes `tests/` a package (adding `__init__.py` files)


pytest_asyncio/plugin.py
```diff
@@ -47,16 +50,14 @@
...
 )
 
-_R = TypeVar("_R")
-
 _ScopeName = Literal["session", "package", "module", "class", "function"]
 _T = TypeVar("_T")
 
 SimpleFixtureFunction = TypeVar(
-    "SimpleFixtureFunction", bound=Callable[..., Awaitable[_R]]
+    "SimpleFixtureFunction", bound=Callable[..., Awaitable[object]]
```
bluetech, line 57: The `_R` here is unbound, I think it doesn't make sense as written.

pytest_asyncio/plugin.py
```diff
@@ -234,8 +238,15 @@ def _preprocess_async_fixtures(
                         f"instead."
                     )
                 )
-            _inject_fixture_argnames(fixturedef, event_loop_fixture_id)
-            _synchronize_async_fixture(fixturedef, event_loop_fixture_id)
+            # TODO: Fix type errors below.
```
bluetech, line 241: The `event_loop_node.stash.get` line can make `event_loop_fixture_id` be `None`, but the functions below don't accept `None`. I'm not sure if:

- This indicates a real problem
- Semantically can't be `None` here (should add an assert)
- The functions should accept `None`

For now I added an ignore which keeps the runtime behavior the same, but let me know if you want another fix.

pytest_asyncio/plugin.py
```diff
@@ -517,20 +528,20 @@ def pytest_pycollect_makeitem_preprocess_async_fixtures(
 @pytest.hookimpl(specname="pytest_pycollect_makeitem", hookwrapper=True)
 def pytest_pycollect_makeitem_convert_async_functions_to_subclass(
     collector: Union[pytest.Module, pytest.Class], name: str, obj: object
-) -> Union[
-    pytest.Item, pytest.Collector, List[Union[pytest.Item, pytest.Collector]], None
-]:
+) -> Generator[None, Any, None]:
```
bluetech, line 531: The `Any` here can be properly typed with newer pytest and/or pluggy but pytest 7.0 is too old for it, so I made it `Any` for now.

seifertm: Thanks for the hint. I opened an issue for dropping pytest 7 support in the future and added a corresponding comment in the code.

pytest_asyncio/plugin.py
seifertm: It shouldn't be possible for `event_loop_fixture_id` to be None here.
I removed the type ignore statements in favour of an `assert`.

seifertm (commented):
This is great! Thank you for the patch.

I'll rebase and add a changelog entry, so we can get this out in a pre-release right away.

Issue #706, opened 2023-12-04
Breaking change in 0.23.*


tkukushkin:
Hello! Something has been broken with the latest pytest-asyncio releases.

Consider such code:

```python
import asyncio
import pytest

@pytest.fixture(scope='session')
def event_loop():
    loop = asyncio.get_event_loop_policy().new_event_loop()
    yield loop
    loop.close()

@pytest.fixture(scope='session')
async def some_async_fixture():
    return asyncio.get_running_loop()

async def test_something(some_async_fixture):
    assert asyncio.get_running_loop() is some_async_fixture
```

pytest.ini:
```ini
[pytest]
asyncio_mode = auto
```

This test passes with pytest-asyncio 0.21.1 but fails with 0.23.0. I am not sure if it's ok, but if it is, IMHO it is breaking change.

albertferras-vrf:
You are right, I forgot about that part

tkukushkin:
@seifertm I'd like to have only one loop for all fixtures and tests, without additional decorators to all tests and fixtures. We have thousands of tests in hundreds of services where all tests and fixtures share one loop and it is crucial for them. Is there any workaround to emulate the old behaviour?

This way https://pytest-asyncio.readthedocs.io/en/v0.23.2/how-to-guides/run_session_tests_in_same_loop.html does not consider fixtures at all(

seifertm:
>  I'd like to have only one loop for all fixtures and tests, without additional decorators to all tests and fixtures.

The linked how-to is supposed to make it easy to add the `asyncio` mark to all of your tests, specifically for large test suites where marking all packages or modules is cumbersome. Unfortunately, there seem to be problems with it, too. Please refer to #705 

> We have thousands of tests in hundreds of services where all tests and fixtures share one loop and it is crucial for them. 

I don't think this use case was considered during the development of v0.22 and v0.23. Can you explain why you need all tests _and_ fixtures to run the same loop? Why

> This way https://pytest-asyncio.readthedocs.io/en/v0.23.2/how-to-guides/run_session_tests_in_same_loop.html does not consider fixtures at all(
That's a fair point. The docs should be updated to explain how fixtures are run.

Fixtures in v0.23 generally behave like tests and choose the event loop of the fixture scope. That means if a fixture has `session` scope it will run in the session-wide loop.

I cannot think of a workaround to switch to the old behaviour at the moment. I suggest pinning pytest-asyncio to <0.23 until this issue is fixed.

[...]

PR #770, opened 2024-01-31
test that mark.asyncio scope is respected

minrk:
test case for tangential issue discovered trying to [provide an example use case](https://github.com/pytest-dev/pytest-asyncio/issues/706#issuecomment-1918982285) for #706.

When a function-scoped async fixture is used, `@pytest.mark.asyncio(scope="module")` on tests is ignored. Notably, when the function-scoped fixture is used in _any_ test, the `scope="module"` is ignored. So `test_module_loop` passes if it is the only test, but if either of the other two tests are enabled, `test_module_loop` will fail without modification. `test_module_loop_function_fixture` always fails.

Draft because this only adds a failing test. I don't think I have the expertise to actually fix the behavior, but hopefully the test case can help someone make sure it's fixed.

### DIFF
--- a/tests/markers/test_module_scope.py
+++ b/tests/markers/test_module_scope.py
@@ -344,3 +344,39 @@ def test_asyncio_mark_handles_missing_event_loop_triggered_by_fixture(
     )
     result = pytester.runpytest("--asyncio-mode=strict")
     result.assert_outcomes(passed=2)
+
+
+def test_asyncio_mark_module_level_loop_reused(
+    pytester: Pytester,
+):
+    pytester.makepyfile(
+        dedent(
+            """\
+            import pytest
+            import asyncio
+            import pytest_asyncio
+
+            @pytest_asyncio.fixture(scope="module")
+            async def module_loop():
+                return asyncio.get_running_loop()
+
+            @pytest_asyncio.fixture(scope="function")
+            async def function_loop():
+                return asyncio.get_running_loop()
+
+            @pytest.mark.asyncio(scope="function")
+            async def test_function_loop(module_loop, function_loop):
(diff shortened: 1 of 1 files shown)


seifertm:
Thanks for the example! I haven't looked at it in detail, yet, but it could be related to #868 .

seifertm:
The reproducer is no longer recent. It comes from a time where *scope* defined both the caching scope of a fixture and its *loop_scope*. I adjusted it to a more recent version of pytest_asyncio:
```python
import pytest
import asyncio
import pytest_asyncio

@pytest_asyncio.fixture(scope="module" loop_scope="module")
async def module_loop():
    return asyncio.get_running_loop()

@pytest_asyncio.fixture(scope="module" loop_scope="function")
async def function_loop():
    return asyncio.get_running_loop()

@pytest.mark.asyncio(loop_scope="function")
async def test_function_loop(module_loop, function_loop):
    assert asyncio.get_running_loop() is function_loop

@pytest.mark.asyncio(loop_scope="module")
async def test_module_loop(module_loop):
    assert asyncio.get_running_loop() is module_loop

@pytest.mark.asyncio(loop_scope="module")
async def test_module_loop_function_fixture(module_loop, function_loop):
    assert asyncio.get_running_loop() is module_loop
```

The test still fails, but the updated version is explicit about the fact that *module_loop* and *function_loop* have a different *loop_scope*. This means they run in different event loops. 

Requesting fixtures with different loop_scope is almost certainly a user error. There are currently no known use cases where this makes sense, so there are plans to turn this into a warning (see #1514).

*Work on this issue was done during the pytest sprint 2026, sponsored by [Omicron](https://www.omicronenergy.com/en/).*

Issue #737, opened 2024-01-02
Support for Pytest 8.0


blink1073:
pytest 8.0.0rc1 was released today, and we're seeing a startup failure with `pytest-asyncio`:

https://github.com/jupyter/jupyter_events/actions/runs/7386335973/job/20092726210

<details>

```
============================= test session starts ==============================
platform linux -- Python 3.11.7, pytest-8.0.0rc1, pluggy-1.3.0 -- /home/runner/.local/share/hatch/env/virtual/jupyter-events/4vpRDyVj/test/bin/python
cachedir: .pytest_cache
rootdir: /home/runner/work/jupyter_events/jupyter_events
configfile: pyproject.toml
testpaths: tests/
plugins: console-scripts-1.4.1, asyncio-0.23.3
asyncio: mode=Mode.AUTO
collecting ... collected 0 items
INTERNALERROR> Traceback (most recent call last):
INTERNALERROR>   File "/home/runner/.local/share/hatch/env/virtual/jupyter-events/4vpRDyVj/test/lib/python3.11/site-packages/_pytest/main.py", line 270, in wrap_session
INTERNALERROR>     session.exitstatus = doit(config, session) or 0
INTERNALERROR>                          ^^^^^^^^^^^^^^^^^^^^^
INTERNALERROR>   File "/home/runner/.local/share/hatch/env/virtual/jupyter-events/4vpRDyVj/test/lib/python3.11/site-packages/_pytest/main.py", line 323, in _main
INTERNALERROR>     config.hook.pytest_collection(session=session)
INTERNALERROR>   File "/home/runner/.local/share/hatch/env/virtual/jupyter-events/4vpRDyVj/test/lib/python3.11/site-packages/pluggy/_hooks.py", line 493, in __call__
INTERNALERROR>     return self._hookexec(self.name, self._hookimpls, kwargs, firstresult)
INTERNALERROR>            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
INTERNALERROR>   File "/home/runner/.local/share/hatch/env/virtual/jupyter-events/4vpRDyVj/test/lib/python3.11/site-packages/pluggy/_manager.py", line 115, in _hookexec
INTERNALERROR>     return self._inner_hookexec(hook_name, methods, kwargs, firstresult)
[...]

PR #776, opened 2024-02-06
fix compatibility with pytest ^8

akeeman:
fixes https://github.com/pytest-dev/pytest-asyncio/issues/737

pytest-asyncio is incompatible with pytest 8.0.0rc1 and 8.0.0rc2, but not with pytest 8.0.0. Although 2 tests had to be updated, as suggested [here](https://github.com/pytest-dev/pytest-asyncio/issues/737#issuecomment-1919889408) by @mgorny.
- test_asyncio_mark_provides_session_scoped_loop_strict_mode: order of module collection has changed from files first to mixed files and folders ([source](https://docs.pytest.org/en/8.0.x/changelog.html#collection-changes))
- test_returns_false_for_unmarked_coroutine_item_in_strict_mode: raises a failure now (again)

### DIFF
--- a/dependencies/default/requirements.txt
+++ b/dependencies/default/requirements.txt
@@ -1,3 +1,3 @@
 # Always adjust install_requires in setup.cfg and pytest-min-requirements.txt
 # when changing runtime dependencies
-pytest >= 7.0.0,<8
+pytest >= 7.0.0, !=8.0.0rc1, !=8.0.0rc2
--- a/dependencies/default/requirements.txt
+++ b/dependencies/default/requirements.txt
@@ -1,3 +1,3 @@
 # Always adjust install_requires in setup.cfg and pytest-min-requirements.txt
 # when changing runtime dependencies
-pytest >= 7.0.0, !=8.0.0rc1, !=8.0.0rc2
+pytest >= 7.0.0,<9
--- a/dependencies/pytest-min/constraints.txt
+++ b/dependencies/pytest-min/constraints.txt
@@ -14,7 +14,7 @@ packaging==23.2
 pluggy==1.3.0
 py==1.11.0
 Pygments==2.16.1
-pytest==8.0.0
+pytest==7.0.0
 requests==2.31.0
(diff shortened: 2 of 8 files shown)


mgorny:
Thanks for doing this!

dependencies/default/requirements.txt
```diff
@@ -1,3 +1,3 @@
 # Always adjust install_requires in setup.cfg and pytest-min-requirements.txt
 # when changing runtime dependencies
-pytest >= 7.0.0,<8
+pytest >= 7.0.0, !=8.0.0rc1, !=8.0.0rc2
```
seifertm, line 3: I ran your patch successfully against pytest-8.0.0rc1 and pytest-8.0.0rc2. This means there's not reason to exclude them in the compatible versions. It's more likely that #764 reduced the number of issues with the collection changes in pytest.

More importantly, we should try to keep the upper version bound at all costs, both in `requirements.txt` and especially in `setup.py`. Otherwise we assume that pytest-asyncio is compatible with potentially breaking changes in pytest 9.

dependencies/pytest-min/constraints.txt
```diff
@@ -14,7 +14,7 @@ packaging==23.2
 pluggy==1.3.0
 py==1.11.0
 Pygments==2.16.1
-pytest==7.0.0
+pytest==8.0.0
```
seifertm, line 17: The intention of the `pytest-min` environment is to test against an older version of pytest. This way, pytest-asyncio doesn't accidentally use any of the newer pytest features without a proper bump of the minimum supported version.

I think we should continue to support pytest 7 for at least a short period of time, in order to allow users to migrate pytest independently from pytest-asyncio.

seifertm (changes requested):
Thanks for the patch!

I think some of the version changes are problematic. Other than that, the PR looks good.

If you don't mind, I'll adjust the version numbers myself and add a changelog entry. Maybe this allows us to get out a pre-release for pytest-asyncio with pytest 8 support today.

PR #800, opened 2024-03-09
Fix compatibility with pytest 8.2 FixtureDef.unittest removal

bluetech:
Since pytest 8.2 (still in development) unittest and regular will behave the same, so unittest special handling is no longer needed.

I also added two unrelated cleanup commits I noticed, if you prefer I can remove them.

### DIFF
--- a/docs/source/reference/changelog.rst
+++ b/docs/source/reference/changelog.rst
@@ -2,6 +2,15 @@
 Changelog
 =========
 
+0.23.6 (2024-03-19)
+===================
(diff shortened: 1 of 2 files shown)


seifertm (approved):
Thank you!

The pytest-asyncio CI only tests against the latest stable release of pytest, so I ran the tests against https://github.com/pytest-dev/pytest/commit/532782a22807c9dd0b0f8c7b1d114bda67f03593 to verify it a second time. Works like a charm. Clean up commits are always welcome, too.

I added a changelog entry and I'll release a new version right away to avoid breakage when pytest 8.2 is released.

Issue #757, opened 2024-01-18
Python 3.11: DeprecationWarning: There is no current event loop


ErikvdVen:
When running the test suite on Python 3.11, tests with a `@pytest.mark.asyncio` decorator, trigger a DeprecationWarning:

```
.../python3.11/site-packages/pytest_asyncio/plugin.py:884: DeprecationWarning: There is no current event loop
   _loop = asyncio.get_event_loop()
```

seifertm:
Which version of pytest-asyncio are you using?
Can you provide a small example that reproduces the issue?

CherrySuryp:
Hi! I encountered same issue with python 3.12

**Here's traceback**
```python
  /python3.12/site-packages/pytest_asyncio/plugin.py:656: DeprecationWarning: There is no current event loop
    old_loop = asyncio.get_event_loop()
```

asl97:
I think this only happens on the older version?

After upgrading from python 3.11.0rc1 to 3.11.7, the warning no longer shows up.
It also doesn't show up on 3.12.1 for me.

```
platform win32 -- Python 3.11.7, pytest-7.4.4, pluggy-1.4.0
platform win32 -- Python 3.12.1, pytest-7.4.4, pluggy-1.4.0
plugins: asyncio-0.23.4
asyncio: mode=Mode.STRICT
```

The warning did shows up on older version of python before I updated it.
> Which version of pytest-asyncio are you using?
```
platform win32 -- Python 3.11.0rc1, pytest-7.4.4, pluggy-1.4.0
plugins: asyncio-0.23.4
asyncio: mode=Mode.STRICT
```

> Can you provide a small example that reproduces the issue?
```python
import pytest

@pytest.mark.asyncio
async def test():
    pass
```

The docs has a footnote on the warning, so it would probably only show up on 3.10.0–3.10.8 and 3.11.0:
https://docs.python.org/3.11/library/asyncio-eventloop.html#asyncio.get_event_loop

[...]

PR #817, opened 2024-04-18
Test for, and suppress, additional get_event_loop() DeprecationWarnings

SyntaxColoring:
This attempts to do what's described in https://github.com/pytest-dev/pytest-asyncio/issues/757#issuecomment-1985912818, to solve continued instances of issue #757:

* ~~Add tests to catch any lingering `asyncio.get_event_loop()` `DeprecationWarning`s.~~
  * ~~The existing tests are already catching a bunch, *but only when they're run under specific Python patch versions.* So the test part of this PR just configures tox and GitHub Actions to do that.~~
  * I could only reproduce the warnings on specific Python 3.10.x patch versions, and I couldn't easily get CI to use those versions, so we don't have tests for this.
* Suppress those warnings.

### DIFF
--- a/tests/async_fixtures/test_async_fixtures_with_finalizer.py
+++ b/tests/async_fixtures/test_async_fixtures_with_finalizer.py
@@ -1,6 +1,5 @@
 import asyncio
 import functools
-import warnings
 
 import pytest
 
@@ -34,11 +33,7 @@ async def port_with_event_loop_finalizer(request):
             # RuntimeError is raised if task is created on a different loop
             await finalizer
 
-        with warnings.catch_warnings():
-            warnings.simplefilter("ignore", DeprecationWarning)
-            loop = asyncio.get_event_loop()
-
-        loop.run_until_complete(port_afinalizer())
+        asyncio.run(port_afinalizer())
 
     worker = asyncio.ensure_future(asyncio.sleep(0.2))
     request.addfinalizer(functools.partial(port_finalizer, worker))
--- a/tests/async_fixtures/test_async_fixtures_with_finalizer.py
+++ b/tests/async_fixtures/test_async_fixtures_with_finalizer.py
@@ -1,5 +1,6 @@
 import asyncio
 import functools
+import warnings
 
 import pytest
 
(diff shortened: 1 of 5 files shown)


tests/async_fixtures/test_async_fixtures_with_finalizer.py
```diff
@@ -33,7 +34,11 @@ async def port_afinalizer():
             # RuntimeError is raised if task is created on a different loop
             await finalizer
 
-        asyncio.get_event_loop().run_until_complete(port_afinalizer())
+        with warnings.catch_warnings():
+            warnings.simplefilter("ignore", DeprecationWarning)
+            loop = asyncio.get_event_loop()
+
+        loop.run_until_complete(port_afinalizer())
```
SyntaxColoring, line 41: I'm not sure about this change.

In all Python versions (at least all the ones I had interpreters installed for), this `get_event_loop()` is implicitly creating an event loop. An event loop never already exists. If you replace this with `get_running_loop()`, it always raises.

I'm not really following what this test is trying to do (what does "port" mean in this context?), but from the comment on line 33, the implicit loop creation doesn't sound intentional.

tox.ini
```diff
@@ -39,6 +47,7 @@ allowlist_externals =
 python =
     3.8: py38, pytest-min
     3.9: py39
+    3.10.8: py3108
```
SyntaxColoring, line 50: Hmm...something's not working with this mapping. The Python 3.10.8 runner uses the `py310` environment, not the `py3108` environment.

It does work locally with pyenv and [tox-pyenv-redux](https://github.com/un-def/tox-pyenv-redux), at least.

seifertm (changes requested):
Thanks a lot for looking into this!

tox.ini
seifertm: Python 3.12 should also emit deprecation warnings, albeit less aggressively. If we cannot reproduce the errors with 3.12, it's fine for me to not have regression tests for the changes.

tests/async_fixtures/test_async_fixtures_with_finalizer.py
SyntaxColoring: Done, TY.

tox.ini
SyntaxColoring: Removed the `p3108` regression test attempts.

SyntaxColoring:
@seifertm Ping—do the latest changes address your feedback?

seifertm (approved):
They do. Thank you for the patch.

I'll aim for a release this week.

PR #863, opened 2024-06-24
Use pytest's FixtureDef type

tamird:


pytest_asyncio/plugin.py
```diff
@@ -61,8 +61,11 @@
 FixtureFunction = Union[SimpleFixtureFunction, FactoryFixtureFunction]
 FixtureFunctionMarker = Callable[[FixtureFunction], FixtureFunction]
 
-# https://github.com/pytest-dev/pytest/pull/9510
-FixtureDef = Any
+# https://github.com/pytest-dev/pytest/commit/fb55615d5e999dd44306596f340036c195428ef1
+if pytest.version_tuple < (8, 0):
```
seifertm, line 65: Perfect work with the version check and continued support for pytest < 8!

seifertm (approved):
Thanks for the contribution.

I don't think this PR needs a changelog entry since the code changes should only affect developers, but not the users. Or am I missing something?

webknjaz:
> I don't think this PR needs a changelog entry since the code changes should only affect developers, but not the users. Or am I missing something?

In pytest itself, I added changelog categories for contributors/downstream folk intended for news/summaries for contributors/core devs. Perhaps, we need to do the same here. But in general, yes, it shouldn't need a change note for the end-users audience.

PR #870, opened 2024-07-05
Automatic updates of copyright year in lib

chirizxc:


webknjaz (changes requested):
That's a incorrect change. Copyright doesn't work like that. It must include the first year of the content creation and optionally years of updates. But it really doesn't make any sense to do such a monkey work with them.

PR #871, opened 2024-07-05
Allow independent caching scope and event loop scopes for async fixtures

### LINKED ISSUE #706: Breaking change in 0.23.*

seifertm:
Pytest-asyncio v0.23 assumes that the event loop scope of a fixture is the same as the caching scope (as in `@pytest.fixture(scope="…")`). This causes problems when fixtures should be cached (e.g. session wide database setup), but the fixture should be run in a differently scoped event loop (e.g. module-scoped loop) or when multiple async fixtures should be used together with different caching scopes.

- [x] Support independent setting of fixture caching scope and its event loop scope (see https://github.com/pytest-dev/pytest-asyncio/issues/706#issuecomment-1913720558)
- [x] Support setting the default loop scope (see https://github.com/pytest-dev/pytest-asyncio/issues/706#issuecomment-1918982285, https://github.com/pytest-dev/pytest-asyncio/issues/706#issuecomment-1923782498)
- [x] ~Deal with mismatching loop scopes of two fixtures used in the same test (see https://github.com/pytest-dev/pytest-asyncio/issues/706#issuecomment-1923782498)~ superseded by #873 
- [x] Deprecate `@pytest.mark.asyncio(scope="…")` in favour of `@pytest.mark.asyncio(loop_scope="…")`
- [x] ~Provide automatic migration tool for the above deprecation?~


seifertm:
I decided not to add an automatic migration tool for the deprecation of `pytest.mark.asyncio(scope=…)`, because a global search and replace should do the trick just as well.

seifertm:
Should we add a migration guide?

PR #873, opened 2024-07-08
Improve detection of multiple event loops being requested in a test

### LINKED ISSUE #868: pytest-asyncio unusable with code using `aiohttp`
It seems pytest-asyncio (`0.23.7`) is currently badly broken with code relying on `aiohttp`.

This simple example currently breaks. The example is heavily simplified from the actual fixture setup. This used to work fine before `0.22`.
```python
import aiohttp
import pytest


@pytest.fixture(scope="session")
async def client(
    # event_loop <- accessing this no longer working. But neither this works without it
):
    async with aiohttp.ClientSession(
        # loop=event_loop <- aiohttp requires using a single event_loop that can not be closed between the tests...
    ) as session:
        yield session


async def test_the_client(client: aiohttp.ClientSession):
    await client.get("http://localhost:8080/foobar")  # RuntimeError: Timeout context manager should be used inside a task
```

Using `asyncio_mode = "auto"`:
```toml
[tool.poetry]
name = "foobar"
package-mode = false

[tool.poetry.dependencies]
python = "~3.12"
aiohttp = "==3.9.5"

[tool.poetry.dev-dependencies]
pytest = "==8.2.2"
pytest-asyncio = "==0.23.7"

[tool.pytest.ini_options]
asyncio_mode = "auto"
```

seifertm:
Fixes #868

### DIFF
--- a/docs/source/reference/changelog.rst
+++ b/docs/source/reference/changelog.rst
@@ -5,6 +5,7 @@ Changelog
 0.23.8 (UNRELEASED)
 ===================
 - Fixes a bug that caused duplicate markers in async tests `#813 <https://github.com/pytest-dev/pytest-asyncio/issues/813>`_
+- Improves detection of multiple event loops being requested by the same test in strict mode `#868 <https://github.com/pytest-dev/pytest-asyncio/issues/868>`_
 
 Known issues
 ------------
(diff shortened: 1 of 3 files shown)


seifertm:
Blocked by #706

PR #886, opened 2024-07-18
Check marker arguments

### LINKED ISSUE #812: Warn when passing scope as positional argument
Recently I used `pytest.mark.asyncio("session")` instead of `pytest.mark.asyncio(scope="session")`.

It took me quite a while to find out why the corresponding test was function scoped instead of session scoped.

It would be nice if either the scope could be passed as positional argument as well, or if there would be a helpful warning or error message in that case, instead of silently swallowing the positional argument.

I can contribute a PR for this feature if you let me know which exact behavior you prefer.

Cito:
This PR implements error checking for the arguments of `mark.asyncio`.

An ValueError with a helpful message is raised when unexpected arguments are passed. 

This should fix issue #812.

### DIFF
--- a/docs/source/reference/changelog.rst
+++ b/docs/source/reference/changelog.rst
@@ -6,6 +6,7 @@ Changelog
 ===================
 - Adds an optional `loop_scope` keyword argument to `pytest.mark.asyncio`. This argument controls which event loop is used to run the marked async test. `#706`_, `#871 <https://github.com/pytest-dev/pytest-asyncio/pull/871>`_
 - Deprecates the optional `scope` keyword argument to `pytest.mark.asyncio` for API consistency with ``pytest_asyncio.fixture``. Users are encouraged to use the `loop_scope` keyword argument, which does exactly the same.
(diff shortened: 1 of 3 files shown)


seifertm (approved):
Thanks for the patch!

I rebased the code to main and adjusted the tests to use the new _loop_scope_ kwargs rather than the deprecated _scope_ kwarg to `@pytest.mark.asyncio`.

I also added a changelog request.

MeggyCal:
As this error is thrown at every unrecognized keyword argument, could it be improved to something more generic, please?

seifertm:
@MeggyCal What problem are you encountering and what improvement do you suggest to solve it?

MeggyCal:
Hi, I've been debugging https://github.com/redis/redis-py/issues/3339 lately. I find `mark.asyncio accepts only a keyword argument 'scope'` cryptic at this point, it would be better to throw `mark.asyncio: Unrecognized keyword argument: %s` or `mark.asyncio does not accept positional arguments. Did you mean to use 'scope'?` depending on the context.

PR #912, opened 2024-08-12
Derive project version using importlib

### LINKED ISSUE #628: Derive project version in docs from Git Tags
The project version specified in `docs/source/conf.py` has not been bumped since v0.20.1.

The `release` field should be derived automatically from the version determined by setuptools_scm, so there's no need to adjust the version manually upon creating a release.

ka28kumar:
Fixes #628 

Using `importlib` to populate the version field in `conf.py`

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -16,3 +16,4 @@ python:
   install:
   - requirements: dependencies/default/constraints.txt
   - requirements: dependencies/docs/constraints.txt
+  - path: .
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -6,10 +6,12 @@
 # -- Project information -----------------------------------------------------
 # https://www.sphinx-doc.org/en/master/usage/configuration.html#project-information
 
+import importlib.metadata
+
 project = "pytest-asyncio"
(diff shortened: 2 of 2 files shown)


seifertm (changes requested):
The doc build in the CI fails with the error `importlib.metadata.PackageNotFoundError: No package metadata was found for pytest-asyncio`. Probably because pytest-asyncio isn't installed in the Python environment used to build the docs.

You could try adding it to the installed packages in `.readthedocs.yaml` and see if it fixes the issue.
```yaml
python:
  install:
    - path: .
…
```

ka28kumar:
Updated `.readthedocs.yaml`.
On a side note, is there a way to validate something like this locally?
I have little experience with ReadTheDocs and I couldn't find anything on this.

seifertm:
> Updated `.readthedocs.yaml`. On a side note, is there a way to validate something like this locally? I have little experience with ReadTheDocs and I couldn't find anything on this.

You're raising a very good point. It's desirable that the process for building the docs in the CI is the same as locally.
At the moment, this is not the case, though. I'll create an issue for that.

seifertm:
Created #917 to track the build issue.

seifertm (approved):

PR #920, opened 2024-08-13
Wire Sphinx builds into the RTD config via tox

### LINKED ISSUE #917: Unify documentation build process for local and CI environment
Currently, the documentation can be built locally using `tox run -e docs` or as part of the CI via readthedocs. However, the build processes are different. This makes it hard to reproduce readthedocs builds locally.

The goal of this issue is to have the same build process for docs, both locally and in the CI.

see https://github.com/pytest-dev/pytest-asyncio/pull/912

webknjaz:
Resolves #917


seifertm (approved):
Great effort, thanks @webknjaz!

I condensed the commit history and adjusted `testpaths` in `setup.cfg` to the new location of the docs.

webknjaz:
Oh, I missed that. This is primarily based on what I did in a number of other places so it wasn't much work, really.

webknjaz:
I wasn't sure about `-b html` because I prefer `-b htmldir` usually, but I figured this is something to be brought up separately so I kept it as is.

PR #941, opened 2024-09-19
Address invisible hidden files in artifact upload action

seifertm:
Coverage writes reports to ".coverage" by default. The GitHub action "upload-artifact" no longer includes hidden files by default. This patch changes the coverage file name to a non-hidden file path.

This fixes the failing CI pipeline for pull requests.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -88,7 +88,7 @@ jobs:
       if: "!endsWith(matrix.os, 'windows')"
       with:
         name: coverage-python-${{ matrix.python-version }}
-        path: .coverage.*
+        path: coverage/coverage.*
         if-no-files-found: error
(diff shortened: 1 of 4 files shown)


Tinche (approved):

Tinche:
Nb: you could have also reenabled hidden file upload in the action with a flag

seifertm:
@Tinche Thanks for the review!

Before this PR, each tox environment created its own coverage file in the project root directory. They don't get overwritten either, so the developer has to clean up the files manually every now and then. I found this a bit annoying.

I considered the `include-hidden-files` option for actions/upload-artifact, but I saw this as a chance to group the coverage files for each tox environment into a single folder to avoid messing up the file explorer for devs :)

PR #968, opened 2024-10-23
Remove support for python3.8

seifertm:
Python 3.8 received its last security release on 2024-09-06 and is now end of life.

see https://peps.python.org/pep-0569/

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -21,7 +21,6 @@ classifiers =
 
   License :: OSI Approved :: Apache Software License
 
-  Programming Language :: Python :: 3.8
   Programming Language :: Python :: 3.9
   Programming Language :: Python :: 3.10
   Programming Language :: Python :: 3.11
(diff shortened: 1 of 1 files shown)


webknjaz (approved):

webknjaz:
It would be nice to have pyupgrade in pre-commit to automate syntax bumps.

Pierre-Sassoulas (commented):
Are you interested in us adding some tooling @seifertm ? (ruff can replace flake8 for example, and it's possible to migrate to pyproject.toml). I would have added the ``UP`` check if ruff was already configured.

seifertm:
> Are you interested in us adding some tooling @seifertm ? (ruff can replace flake8 for example, and it's possible to migrate to pyproject.toml). I would have added the `UP` check if ruff was already configured.

Absolutely. I'm very open to these tooling updates.
I use ruff in other projects and it works just fine. Pyupgrade is also an option.

PR #969, opened 2024-10-24
Migrate from setup.cfg remvoe flake8, add ruff, pyupgrade pydocstyle

Pierre-Sassoulas:
Follow-up to #968


pyproject.toml
```diff
@@ -68,10 +68,44 @@ write_to = "pytest_asyncio/_version.py"
...
 [tool.ruff]
 line-length = 88
+format.docstring-code-format = true
 lint.select = [
+  "B", # bugbear
+  "D", # pydocstyle
   "E", # pycodestyle
   "F", # pyflakes
-  "W", # pycodestyle
+  #  "FA100",  # add future annotations (not automated but would permit to update 'Union' via UP)
```
Pierre-Sassoulas, line 77: There's probably a tool that would allow us to not have to copy paste ``from __future__ import annotations`` to the top of 30 files 😄

Pierre-Sassoulas: There was indeed

docs/concepts_function_scope_example.py
```diff
@@ -1,3 +1,5 @@
+from __future__ import annotations
```
seifertm, line 1: I don't care if we make this import for test files, but I don't think we should import this in the files in _docs._
The Python scripts in the _docs_ folder get included as snippets in the documentation. The future import adds noise to the snippet and may be confusing.

seifertm (changes requested):

docs/concepts_function_scope_example.py
Pierre-Sassoulas:
```suggestion
```

Pierre-Sassoulas: Right, fixed :D

seifertm:
Thanks, merged!

PR #970, opened 2024-10-24
Migrate from flake8 to ruff

Pierre-Sassoulas:
Saner review and CI debugging than in #969


.pre-commit-config.yaml
```diff
@@ -68,6 +68,12 @@ repos:
   rev: 0.29.3
   hooks:
   - id: check-github-actions
+- repo: https://github.com/tox-dev/pyproject-fmt
```
seifertm, line 71: I wasn't aware of this hook. Nice addition!

.pre-commit-config.yaml
```diff
@@ -68,6 +68,12 @@ repos:
...
+  rev: 2.3.1
+  hooks:
+  - id: pyproject-fmt
+      # https://pyproject-fmt.readthedocs.io/en/latest/#calculating-max-supported-python-version
+    additional_dependencies: [tox>=4.9]
```
seifertm, line 76: It probably makes sense to also bump the `minversion` in _tox.ini_ accordingly.

seifertm (approved):
Works smoothly in my setup, as expected. Thanks!

.pre-commit-config.yaml
Pierre-Sassoulas: Good catch, done in [20a1097](https://github.com/pytest-dev/pytest-asyncio/pull/970/commits/20a1097063df1f4d49eadb0be5dbc664ff4bed0e)

PR #971, opened 2024-10-24
Migrate from setup.cfg to pyproject.toml

Pierre-Sassoulas:
Follow-up to #970


Pierre-Sassoulas:
The error can be reproduced locally with
```
pip install build
python -m build
```

Pierre-Sassoulas:
I've done the lazy fix here, I'm not using setuptool_scm a lot myself. I think a partial migration is still better than none at all.

seifertm (approved):
I agree. We probably have to revisit the setuptools_scm configuration for this to resolve, but I don't think it's worth blocking on that.

PR #990, opened 2024-11-18
display full param name asyncio_default_fixture_loop_scope

MarkusBiggus:
confirmed verbose output displays correct name:

asyncio: mode=Mode.AUTO, asyncio_default_fixture_loop_scope=function

file __version__ not present and I don't understand the issue described in setup.cfg

guessed version in changelog.rst with totally random date (my birthday)


seifertm (approved):
Apologies for leaving this open for such a long time. The change absolutely makes sense.

I rebased the branch, addressed the conflict, and squashed the commit that was added by "pre-commit ci".

With regards to your questions:
For new versions, we usually use the date "UNRELEASED".
The `_version.py` file is generated by setuptools_scm when packaging the plugin into an sdist/Python wheel. Setuptools_scm derives the project version from the tags of the Git repository. That means we can create new releases by creating an appropriate Git tag.

These things should be mentioned in a contribution guide and/or the documentation, so thanks for pointing out the deficiencies :)

PR #1004, opened 2024-12-02
Rename 'scope' to 'loop_scope' in error message

tusharsadhwani:
Ensures that when a wrong argument is passed, the non-deprecated `loop_scope` keyword arg is recommended.


seifertm (changes requested):
Thanks for pointing this out!
Would you be willing to adjust the tests accordingly and add an entry to the changelog? Or would you prefer one of the maintainers take care of it?

tusharsadhwani:
I can do that. Let me fix the tests too.

tusharsadhwani:
@seifertm how's this?

seifertm (approved):
Perfect!
I squashed the commits and we can merge.

Thanks for the contribution!

PR #1008, opened 2024-12-06
Maintain contextvars between fixtures and tests

bcmills:
The approach I've taken here is to inspect the context after running the fixture setup task and copy any changed values out into the test's ambient context. When we run the finalizer to tear down the fixture, we then reset the modified variables.

Some limitations:
- This requires Python 3.11 or higher (for https://github.com/python/cpython/issues/91150).
- Because we have to iterate over the fixture context to copy out changes, this approach may be slow for fixtures that run in contexts with a large number of variables.

Fixes pytest-dev/pytest-asyncio#127.

### DIFF
--- a/pytest_asyncio/plugin.py
+++ b/pytest_asyncio/plugin.py
@@ -381,12 +381,10 @@ def _create_task_in_context(loop, coro, context):
     the API added for https://github.com/python/cpython/issues/91150.
     On earlier versions, the returned task will use the default context instead.
     """
-    if context is not None:
-        try:
-            return loop.create_task(coro, context=context)
-        except TypeError:
-            pass
-    return loop.create_task(coro)
+    try:
+        return loop.create_task(coro, context=context)
+    except TypeError:
+        return loop.create_task(coro)
 
 
 def _wrap_async_fixture(fixturedef: FixtureDef) -> None:
--- a/pytest_asyncio/plugin.py
+++ b/pytest_asyncio/plugin.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 import asyncio
 import contextlib
+import contextvars
 import enum
 import functools
 import inspect
@@ -318,6 +319,8 @@ def _wrap_asyncgen_fixture(fixturedef: FixtureDef) -> None:
         kwargs.pop(event_loop_fixture_id, None)
         gen_obj = func(**_add_kwargs(func, kwargs, event_loop, request))
 
+        context = _event_loop_context.get(None)
+
         async def setup():
             res = await gen_obj.__anext__()  # type: ignore[union-attr]
             return res
@@ -335,9 +338,11 @@ def _wrap_asyncgen_fixture(fixturedef: FixtureDef) -> None:
                     msg += "Yield only once."
                     raise ValueError(msg)
 
-            event_loop.run_until_complete(async_finalizer())
+            task = _create_task_in_context(event_loop, async_finalizer(), context)
+            event_loop.run_until_complete(task)
 
-        result = event_loop.run_until_complete(setup())
+        setup_task = _create_task_in_context(event_loop, setup(), context)
+        result = event_loop.run_until_complete(setup_task)
         request.addfinalizer(finalizer)
         return result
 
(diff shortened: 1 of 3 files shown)


bcmills:
> - If a synchronous fixture runs after the corresponding event loop has already been created, and it modifies context vars in the thread's current context, those changes will not propagate into async fixtures and functions.

It occurs to me that we can get rid of this caveat by copying the modified values out to the ambient context after running the setup phase, and resetting them after the finalizer. I'll look into that approach.

bcmills:
> It occurs to me that we can get rid of this caveat by copying the modified values out to the ambient context after running the setup phase, and resetting them after the finalizer.

Done, and I'm even happier with this approach!

pytest_asyncio/plugin.py
```diff
@@ -360,6 +403,10 @@ async def setup():
             res = await func(**_add_kwargs(func, kwargs, event_loop, request))
             return res
 
+        # Since the fixture doesn't have a cleanup phase, if it set any context
```
bcmills, line 406: I'm not quite sure about this part. Would it be better to copy in the variable changes and add a finalizer to reset them, as we do for generator fixtures? That would also be fairly intuitive, but the behavior of async fixtures would still be different than synchronous fixture functions.

bcmills: On further consideration, I think it's better to copy in the changes. If a generator fixture propagates contextvar changes, it seems surprising for a non-generator fixture to fail to do so.

And if changes _are_ copied in, I would rather reset on teardown than just leave them in place: creating an `asyncio.Task` usually copies the context, so it shouldn't be terribly surprising for contextvar changes to eventually go out of scope, and that reduces the risk of unexpected crosstalk between tests.

tests/async_fixtures/test_async_fixtures_contextvars.py
```diff
@@ -0,0 +1,73 @@
...
+@contextmanager
+def context_var_manager(value):
+    token = _context_var.set(value)
+    try:
+        yield
+    finally:
+        _context_var.reset(token)
+
+
+@pytest.fixture(scope="function")
```
seifertm, line 26: The test cases are well thought out! I took me some time to understand them, though. I think this is due to the fact that the test cases build upon each other.

Most of the other test files have been refactored to use _pytest.Pytester_ (refer to [this test](https://github.com/pytest-dev/pytest-asyncio/blob/c4c1b482de0472d91b40ba0c07935d704c483bce/tests/markers/test_function_scope.py) as an example). Pytester is specifically designed to test pytest plugins. Among other things, this allows isolation of tests against each other, e.g. avoiding accidentally modified context in subsequent tests.

In my opinion, the new tests should make use of Pytester. I also think the tests should be grouped according to their test intention. That will probably create some code duplication in the tests, but I think the benefits outweigh this drawback.

For example:
```python
def test_context_var_in_requesting_fixture_does_not_affect_parent_fixture(pytester: Pytester):
    pytester.makepyfile(
      dedent(
          """\
          async def no_var_fixture():
              with pytest.raises(LookupError):
                  _context_var.get()
              yield
              with pytest.raises(LookupError):
                  _context_var.get()

          @pytest_asyncio.fixture(scope="function")
          async def var_fixture_1(no_var_fixture):
              with context_var_manager("value1"):
                yield

          async def test_request_fixture(var_fixture_1):
              pass
          """
      )
    result = pytester.runpytest("--asyncio-mode=strict")  # Use strict mode, if the test isn't specific to auto mode.
    result.assert_outcomes(passed=2)
 ```

Could you give refactoring the tests a try?

**Review by seifertm — CHANGES_REQUESTED** [review]
You put a lot of effort into this, thanks! I find this a fairly elegant way to address the long-standing problem with context vars in pytest-asyncio. Your use of helper functions makes the code easy to follow and there are tests to support the change.

I'd appreciate if:
1. You addressed my comment regarding the structure of the tests.
2. You added a corresponding entry in `docs/reference/changelog.rst` that mentions the support for contextvars and the limitation to Python >=3.11.

File: tests/async_fixtures/test_async_fixtures_contextvars.py (hunk 2 above)
Comment (bcmills) [reply tests/async_fixtures/test_async_fixtures_contextvars.py]: Done. There is indeed more duplication, but the purpose of each part of the test should be clearer now.

File: pytest_asyncio/plugin.py [hunk 3]
```diff
@@ -385,6 +410,57 @@ def _get_event_loop_fixture_id_for_async_fixture(
     return event_loop_fixture_id
 
 
+def _create_task_in_context(loop, coro, context):
```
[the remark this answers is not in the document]
bcmills: Done. The `coro` parameter needed to be` Coroutine` specifically rather than `Awaitable` in general due to the signature of `loop.create_task`.

bcmills:
Thanks for the review! I think I've addressed all of your comments, but please let me know if I missed something.

seifertm (approved):
That's exactly what I had in mind, thanks!

PR #1009, opened 2024-12-09
docs: describe when UsageError is raised for loop_scope="class" marker

zed:
before that: docs claimed (erroneously) the opposite of the actual (and intended) behavior.


seifertm (approved):
Absolutely! Thanks for not just pointing out the mistake, but also fixing it right away.

PR #1014, opened 2024-12-11
Mention `asyncio_default_fixture_loop_scope` in v0.24 changelog

robsdedude:
This config option is arguably a new API surface, and it took me a fair bit of git archaeology to figure out from which version on I can use this config option.

This entry might save some users some time in the future :)


seifertm (approved):
Thanks for the PR.

We should probably have done a better job advertising the new configuration option. Adding it to the changelog seems fair.

PR #1029, opened 2024-12-30
Fix wrong event loop reference with mixed scoped tests

### LINKED ISSUE #950: Requesting loop_scope="module" (or "session" etc.) returns incorrect event_loop if another test function requests a different loop.
First of all, thanks for the great pytest plugin! 
Also, I like the idea of being able to specify event_loop scopes explicitly and separately to the fixture scopes. 

While playing with this, I kept running into issues that my "session" scoped event_loop seemed to change between tests that were running in different files, but only if certain other tests were run... 

I think this example reproduces the issues in a single file.  I recognize that the docs explicitly advise against doing this, but I hope it makes the problem easier to reason about.

Here's the example.

```python
import pytest_asyncio
import asyncio
import pytest


class FakeAsyncConnection:
    def __init__(self, loop):
        self.loop = loop

    async def do_something(self):
        # Check if the current loop is the same as the one with which the
        #  connection was created
        if asyncio.get_event_loop() is not self.loop:
            raise RuntimeError(
                "This connection is being used with a different event loop!")
        return "Success"


@pytest_asyncio.fixture(scope="module", loop_scope="module")
async def async_connection():
    """Set up a async connection object with module scope."""
    event_loop = asyncio.get_event_loop()
    print(f"Setting up fixture: event_loop_id {id(event_loop)}")
    connection = FakeAsyncConnection(event_loop)
    yield connection


@pytest.mark.asyncio(loop_scope="module")
async def test_use_module_scope_loop_1(async_connection):
    """Use module loop"""
    print(f"Test using loop with id: {id(asyncio.get_event_loop())}")
    result = await async_connection.do_something()
    assert result == "Success"


@pytest.mark.asyncio(loop_scope="module")
async def test_use_module_scope_loop_2(async_connection):
    """Use module loop again"""
    print(f"Test using loop with id: {id(asyncio.get_event_loop())}")
    result = await async_connection.do_something()
    assert result == "Success"


@pytest.mark.asyncio(loop_scope="function")
async def test_use_function_scope_loop_1(async_connection):
    """Use function loop"""
    print(f"Test using loop with id: {id(asyncio.get_event_loop())}")
    with pytest.raises(RuntimeError, match="This connection is being used with a different event loop!"):
        # This should raise an error because the connection is being used with a different loop
        await async_connection.do_something()


@pytest.mark.asyncio(loop_scope="module")
async def test_use_module_scope_loop_3(async_connection):
    """Unexpectedly fail to use module scope again"""
    print(f"Test using loop with id: {id(asyncio.get_event_loop())}")
    result = await async_connection.do_something()
    assert result == "Success"
```

I would expect all tests to pass, however, the final test `test_use_module_scope_loop_3` fails **only** if the `test_use_function_scope_loop_1` is present. If the function scope one is commented out, the final test does pass (as expected).

The fixtures aren't obviously set up incorrectly (running with `--setup-show`):
```
SETUP    S event_loop_policy
    SETUP    M tests/test_a.py::<event_loop> (fixtures used: event_loop_policy)
    SETUP    M async_connection
        tests/test_a.py::test_use_module_scope_loop_1 (fixtures used: async_connection, event_loop_policy, request, tests/test_a.py::<event_loop>).
        tests/test_a.py::test_use_module_scope_loop_2 (fixtures used: async_connection, event_loop_policy, request, tests/test_a.py::<event_loop>).
        SETUP    F event_loop
        tests/test_a.py::test_use_function_scope_loop_1 (fixtures used: async_connection, event_loop, event_loop_policy, request).
        TEARDOWN F event_loop
        tests/test_a.py::test_use_module_scope_loop_3 (fixtures used: async_connection, event_loop_policy, request, tests/test_a.py::<event_loop>)F
    TEARDOWN M async_connection
    TEARDOWN M tests/test_a.py::<event_loop>
TEARDOWN S event_loop_policy
```

But for some reason I don't understand, the module scoped event loop changes for the last test.

The printed loop ids tell the same story... The fixture and first two tests all get the the same loop_id as expected. The function scope test gets a new one as expected. Then the final module scope test also gets a new loop_id (different from both previous loop_ids) unexpectedly. 

Versions:
python: 3.12.7
pytest: 8.3.3
pytest-asyncio: 0.24.0

Also, my pytest settings are:
```
[tool.pytest.ini_options]
asyncio_mode = "auto"
asyncio_default_fixture_loop_scope="function"
```

seifertm:
Issue #950 describes a problem with tests that have mixed event loop scopes. When a function-scoped loop is executed in between two tests with wider loop scope, the tests after the function-scoped tests reference the wrong event loop.

This is caused by the implementation of the _event_loop_ fixture and its finalizers. This PR addresses the issue twofold:
1. It changes the implementation of the _event_loop_ fixture so that it properly restores the current event loop policy and event loop when the fixture is torn down.
2. It changes the _event_loop_ fixture finalizers to only close loops that were not created by pytest-asyncio. This prevents the finalizers from accidentally closing a module-scoped loop provided by the plugin, for example.

Fixes #950

### DIFF
--- a/docs/reference/changelog.rst
+++ b/docs/reference/changelog.rst
@@ -2,6 +2,11 @@
 Changelog
 =========
 
+0.25.1 (UNRELEASED)
+===================
+- Fixes an issue that caused a broken event loop when a function-scoped test was executed in between two tests with wider loop scope `#950 <https://github.com/pytest-dev/pytest-asyncio/issues/950>`_
+
+
 0.25.0 (2024-12-13)
 ===================
 - Deprecated: Added warning when asyncio test requests async ``@pytest.fixture`` in strict mode. This will become an error in a future version of flake8-asyncio. `#979 <https://github.com/pytest-dev/pytest-asyncio/pull/979>`_
--- a/pytest_asyncio/plugin.py
+++ b/pytest_asyncio/plugin.py
@@ -861,7 +861,7 @@ def pytest_fixture_setup(
         # Weird behavior was observed when checking for an attribute of FixtureDef.func
         # Instead, we now check for a special attribute of the returned event loop
         fixture_filename = inspect.getsourcefile(fixturedef.func)
-        if not getattr(loop, "__original_fixture_loop", False):
+        if not _is_pytest_asyncio_loop(loop):
             _, fixture_line_number = inspect.getsourcelines(fixturedef.func)
             warnings.warn(
                 _REDEFINED_EVENT_LOOP_FIXTURE_WARNING
@@ -1154,7 +1154,7 @@ def event_loop(request: FixtureRequest) -> Iterator[asyncio.AbstractEventLoop]:
         # set this value.
         # The magic value must be set as part of the function definition, because pytest
         # seems to have multiple instances of the same FixtureDef or fixture function
-        loop.__original_fixture_loop = True  # type: ignore[attr-defined]
+        loop = _make_pytest_asyncio_loop(loop)
         yield loop
         loop.close()
 
--- a/pytest_asyncio/plugin.py
+++ b/pytest_asyncio/plugin.py
@@ -709,8 +709,7 @@ def pytest_collectstart(collector: pytest.Collector) -> None:
     ) -> Iterator[asyncio.AbstractEventLoop]:
         new_loop_policy = event_loop_policy
         with _temporary_event_loop_policy(new_loop_policy):
-            loop = asyncio.new_event_loop()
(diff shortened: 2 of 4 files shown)


asvetlov (approved):
Looks good.
The code is extremely complex now; I have no idea how to simplify it and keep backward compatibility, though.

seifertm:
Thanks for having a look!

I agree that the code is really complex. It takes hours to debug even simple changes and understand 

We really made an effort to keep backwards compatibility, so that made the situation worse. Maybe we should start following through with all the deprecations that were introduced and start cleaning up the code.

PR #1034, opened 2025-01-05
Call `loop.shutdown_asyncgens` before closing the event loop

provinzkraut:
#222 was closed, mentioning the issues being resolved in the yet-to-be-done #235, running with `-W error` in Python 3.13, currently fails when there are unclosed async generators. 

This is a minimally invasive fix, without attempting to behave like `asyncio.run` on shutdown in all other scenarios.

### DIFF
--- a/docs/reference/changelog.rst
+++ b/docs/reference/changelog.rst
@@ -2,6 +2,11 @@
 Changelog
 =========
 
+0.25.2 (UNRELEASED)
+===================
+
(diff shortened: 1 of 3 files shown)


seifertm (approved):
Solid patch, thank you!

Can you please add an entry to _docs/reference/changelog.rst_ as well, where you briefly explain the change and link to this PR? We can create another patch release for this. Just set the date to _UNRELEASED_ and we'll handle the rest.

provinzkraut:
@seifertm Added a comment!

seifertm:
Thanks for the contribution!

seifertm:
Squashed the commits and extended commit message.

PR #1043, opened 2025-01-14
docs: specify how asyncio's mode can be set

JP-Ellis:
When reading the docs, it wasn't clear to me how to set asyncio's mode and I had to rely on looking at the source code. I thought it would be nice to expand the docs to be more explicit in this regard.

### DIFF
--- a/docs/concepts.rst
+++ b/docs/concepts.rst
@@ -47,8 +47,12 @@ Assigning neighboring tests to different event loop scopes is discouraged as it 
 Test discovery modes
 ====================
 
-Pytest-asyncio provides two modes for test discovery, *strict* and *auto*.
+Pytest-asyncio provides two modes for test discovery, *strict* and *auto*. This can be set through Pytest's ``--asyncio-mode`` command line flag, or through the configuration file:
 
+.. code-block:: toml
+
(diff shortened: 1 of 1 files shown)


asvetlov (approved):
lgtm

seifertm:
> When reading the docs, it wasn't clear to me how to set asyncio's mode and I had to rely on looking at the source code. I thought it would be nice to expand the docs to be more explicit in this regard.

That's a good point, thanks for raising it! I can relate to the problem.

The pytest-asyncio documentation roughly follows the [Diataxis](https://diataxis.fr/) approach at the moment.
The _Concepts_ section (called _Explanation_ in Diataxis) is only concerned with providing information to understand pytest-asyncio, not necessarily how to use it. There's also a [configuration reference for asyncio_mode](https://pytest-asyncio.readthedocs.io/en/latest/reference/configuration.html#asyncio-mode) which should contain all relevant information.

Personally, I would prefer if you:
- Link to the asyncio_mode reference from _concepts.rst_ 
- and/or create a new "How to configure asyncio_mode" in the how-tos.

seifertm:
I guess it's better to have this.
Thanks @JP-Ellis !

PR #1045, opened 2025-01-18
Fix FixtureFunction TypeAlias and TypeVars

cdce8p:
If TypeVars are used in TypeAliases, they need to be passed along as well. Otherwise it's just interpreted as `Any` similar to all other generic type without subscripts.

This PR reworks the TypeAliases and TypeVars to properly type the `fixture` function.

### DIFF
--- a/docs/reference/changelog.rst
+++ b/docs/reference/changelog.rst
@@ -5,6 +5,8 @@ Changelog
 0.26.0 (UNRELEASED)
 ===================
 - Adds configuration option that sets default event loop scope for all testss `#793 <https://github.com/pytest-dev/pytest-asyncio/issues/793>`_
+- Improved type annotations for ``pytest_asyncio.fixture`` `#1045 <https://github.com/pytest-dev/pytest-asyncio/pull/1045>`_
+- Added ``typing-extensions`` as additional dependency for Python ``<3.10`` `#1045 <https://github.com/pytest-dev/pytest-asyncio/pull/1045>`_
 
 
 0.25.2 (2025-01-08)
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -38,6 +38,7 @@ dynamic = [
 
 dependencies = [
   "pytest>=8.2,<9",
+  "typing-extensions>=4.12; python_version<'3.10'",
 ]
 optional-dependencies.docs = [
   "sphinx>=5.3",
--- a/pytest_asyncio/plugin.py
+++ b/pytest_asyncio/plugin.py
@@ -56,14 +56,18 @@ from pytest import (
 _ScopeName = Literal["session", "package", "module", "class", "function"]
 _T = TypeVar("_T")
 
-SimpleFixtureFunction = TypeVar(
-    "SimpleFixtureFunction", bound=Callable[..., Awaitable[object]]
+FixtureFunctionT = TypeVar(
+    "FixtureFunctionT",
+    bound=Union[Callable[..., Awaitable[Any]], Callable[..., AsyncIterator[Any]]],
 )
-FactoryFixtureFunction = TypeVar(
-    "FactoryFixtureFunction", bound=Callable[..., AsyncIterator[object]]
+FixtureFunctionU = TypeVar(
+    "FixtureFunctionU",
+    bound=Union[Callable[..., Awaitable[Any]], Callable[..., AsyncIterator[Any]]],
 )
-FixtureFunction = Union[SimpleFixtureFunction, FactoryFixtureFunction]
-FixtureFunctionMarker = Callable[[FixtureFunction], FixtureFunction]
+FixtureFunctionType = Union[
+    Callable[..., Awaitable[Any]], Callable[..., AsyncIterator[Any]]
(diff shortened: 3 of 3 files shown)


seifertm (commented):
Thanks for the patch.
I must admit that I don't fully understand it, though.

You're saying that TypeVars need used as part of a function signature. Previously, they were used in type aliases which, in turn, were used in the function signature, but the type checker doesn't understand that indirection. Is that the problem that this patch solves?

cdce8p:
> You're saying that TypeVars need used as part of a function signature. Previously, they were used in type aliases which, in turn, were used in the function signature, but the type checker doesn't understand that indirection. Is that the problem that this patch solves?

The issue here are the generic type aliases. Let me provide a simplified example:
```py
T = TypeVar("T")
U = TypeVar("U")
Alias = Union[T, U]
x: Alias
reveal_type(x)  # Union[Any, Any]
```

Alias here is similar to `FixtureFunction`. To be used correctly, the type alias itself would need to change to `Alias[T, U]` which the type checker could infer as `Union[T?, U?]`. The mypy docs on that topic might also be helpful: https://mypy.readthedocs.io/en/stable/generics.html#generic-type-aliases

--
To come back to the current usage. If you run pyright (in strict mode), these are a few of the error it generates
```py
  /.../pytest_asyncio/plugin.py:66:1 - error: Type of "FixtureFunctionMarker" is partially unknown
    Type of "FixtureFunctionMarker" is "type[(Unknown) -> Unknown]" (reportUnknownVariableType)
  /.../pytest_asyncio/plugin.py:66:35 - error: Expected type arguments for generic type alias "FixtureFunction" (reportMissingTypeArgument)
  /.../pytest_asyncio/plugin.py:66:53 - error: Expected type arguments for generic type alias "FixtureFunction" (reportMissingTypeArgument)
  /.../pytest_asyncio/plugin.py:119:5 - error: Return type is unknown (reportUnknownParameterType)
  /.../pytest_asyncio/plugin.py:120:5 - error: Type of parameter "fixture_function" is unknown (reportUnknownParameterType)
  /.../pytest_asyncio/plugin.py:120:23 - error: Expected type arguments for generic type alias "FixtureFunction" (reportMissingTypeArgument)
  /.../pytest_asyncio/plugin.py:132:6 - error: Expected type arguments for generic type alias "FixtureFunction" (reportMissingTypeArgument)
  /.../pytest_asyncio/plugin.py:136:5 - error: Return type, "(Unknown) -> Unknown", is partially unknown (reportUnknownParameterType)
  /.../pytest_asyncio/plugin.py:152:5 - error: Return type, "Unknown | ((Unknown) -> Unknown)", is partially unknown (reportUnknownParameterType)
  /.../pytest_asyncio/plugin.py:153:5 - error: Type of parameter "fixture_function" is partially unknown
    Parameter type is "Unknown | None" (reportUnknownParameterType)
  /.../pytest_asyncio/plugin.py:153:23 - error: Expected type arguments for generic type alias "FixtureFunction" (reportMissingTypeArgument)
  /.../pytest_asyncio/plugin.py:156:6 - error: Expected type arguments for generic type alias "FixtureFunction" (reportMissingTypeArgument)
```

As you might see, there are a lot of `Unknown` types which are a result of the issue mentioned above. In the end it does work out though (at least for pyright) as it simply disregards the decorator as "untyped". Mypy interprets the return type as `Any`.
```py
import pytest_asyncio

@pytest_asyncio.fixture()  # Overload 1
async def fixture_1() -> AsyncGenerator[int]:
    yield 1

@pytest_asyncio.fixture  # Overload 2
async def fixture_2() -> AsyncGenerator[int]:
    yield 1
```
```py
# pyright (the decorator is just ignored)
reveal_type(fixture_1)  # () -> AsyncGenerator[int, None]
reveal_type(fixture_2)  # () -> AsyncGenerator[int, None]

# mypy
reveal_type(fixture_1)  # Union[Any, Any]
reveal_type(fixture_2)  # Union[Any, Any]
```

After the changes in this PR, pyright no longer emits any warnings for the `pytest_asyncio.fixture` decorator. The reveal_type outputs also look better
```py
# pyright (decorator inferred correctly)
reveal_type(fixture_1)  # () -> AsyncGenerator[int, None]
reveal_type(fixture_2)  # () -> AsyncGenerator[int, None]

# mypy
reveal_type(fixture_1)  # Never  # probably a bug in mypy
reveal_type(fixture_2)  # def () -> typing.AsyncGenerator[builtins.int, None]
```

seifertm (changes requested):
Thanks for the explanation and the overhaul! I like the second version with ParamSpec much better.

I can imagine it was a lot of effort to get to this point, especially with different behaviors between mypy and pyright.

Before we can merge, we need to inform the users about the change. Could you add an entry to `docs/reference/changelog.rst` where you briefly describe what's been fixed? Please also mention the added dependency on typing_extensions in a separate bullet point, since this is relevant for downstream packagers.

seifertm (approved):

seifertm:
Thanks @cdce8p!

PR #1051, opened 2025-01-27
Avoid errors in cleanup when event loop is already closed

### LINKED ISSUE #1040: Regression in 0.25.2 - Event loop is closed
If there is a user defined fixture which handles loop shutdown and closing the loop on its own, the plugin raises an error.

Expected behavior: If user code closes the event loop, pytest-asyncio should handle this transparently.

Actual behavior: pytest-asyncio raises RuntimeError: Event loop is closed.

```
..\..\AppData\Local\Programs\Python\Python310\lib\contextlib.py:142: in __exit__
    next(self.gen)
.venv\lib\site-packages\pytest_asyncio\plugin.py:1168: in _provide_event_loop
    loop.run_until_complete(loop.shutdown_asyncgens())
..\..\AppData\Local\Programs\Python\Python310\lib\asyncio\base_events.py:624: in run_until_complete
    self._check_closed()
```

Example working with 0.25.1, crashes with 0.25.2:

```python

import asyncio

import pytest


@pytest.fixture(autouse=True)
def handle_event_loop(request: pytest.FixtureRequest):
    event_loop = None
    if "event_loop" in request.fixturenames:
        event_loop = request.getfixturevalue("event_loop")

    yield

    if event_loop is None or event_loop.is_closed():
        return
    try:
        asyncio.runners._cancel_all_tasks(event_loop)
        event_loop.run_until_complete(event_loop.shutdown_asyncgens())
        if hasattr(event_loop, "shutdown_default_executor"):  # Python 3.9+
            event_loop.run_until_complete(event_loop.shutdown_default_executor())
    finally:
        event_loop.close()


async def test():
    assert True

```

minrk:
check `is_closed()` before calling cleanup methods and catch errors in cleanup to avoid problems

closes #1040

### DIFF
--- a/docs/reference/changelog.rst
+++ b/docs/reference/changelog.rst
@@ -9,9 +9,15 @@ Changelog
 - Added ``typing-extensions`` as additional dependency for Python ``<3.10`` `#1045 <https://github.com/pytest-dev/pytest-asyncio/pull/1045>`_
 
 
+0.25.3 (2025-01-28)
+===================
+- Avoid errors in cleanup of async generators when event loop is already closed `#1040 <https://github.com/pytest-dev/pytest-asyncio/issues/1040>`_
+
+
 0.25.2 (2025-01-08)
 ===================
 - Call ``loop.shutdown_asyncgens()`` before closing the event loop to ensure async generators are closed in the same manner as ``asyncio.run`` does `#1034 <https://github.com/pytest-dev/pytest-asyncio/pull/1034>`_
+
 
 0.25.1 (2025-01-02)
 ===================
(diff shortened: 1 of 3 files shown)


minrk:
Not sure why codecov misreports coverage. A local run shows the missed lines are actually covered by 8bc0e19

seifertm (approved):
Thanks a lot for the bug fix!

I agree that downgrading exceptions during cleanup to a warning makes sense, as outlined in my comment on the bug report.

I took the liberty to modify the changelog commit history for easier cherry-picking into the v0.25 release branch. I think we should do a separate bugfix release for v0.25, rather than adding this to the v0.26 release.

minrk:
Thanks!

PR #1106, opened 2025-05-02
Remove event_loop fixture

seifertm:
This PR removes the deprecated *event_loop* fixture along with the associated warnings, such as the unclosed loop warning, the warning about requesting *event_loop* in async fixtures and tests, and the warning about multiple loops being requested in the same test.

This patch follows up the deprecation of the *event_loop* fixture in v0.22/v0.23 and represents a major cleanup of the pytest-asyncio code base.


Dreamsorcerer:
Hmm, we've got some rather awkward tests going on...

We have things like:
```
def test_one_thousand_round_trip_websocket_text_messages(
    event_loop: asyncio.AbstractEventLoop,
    aiohttp_client: AiohttpClient,
    benchmark: BenchmarkFixture,
) -> None:
    ...

    @benchmark
    def _run() -> None:
        event_loop.run_until_complete(run_websocket_benchmark())
```

But, aiohttp_client fixture is async, so it must be running in a loop context. So, if I create a new loop or something in this test, then I get problems like:
```
RuntimeError: Task <Task pending name='Task-34' coro=<RequestHandler.shutdown() running at /Users/runner/work/aiohttp/aiohttp/aiohttp/web_protocol.py:336> cb=[gather.<locals>._done_callback() at /Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/asyncio/tasks.py:764]> got Future <Future pending> attached to a different loop
```

PR #1120, opened 2025-05-14
Fix release note extraction from Git tag annotation

### LINKED ISSUE #1113: Empty release notes when extracting from Git tag

seifertm:
Closes #1113

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -155,7 +155,7 @@ jobs:
       if: github.event_name == 'push' && contains(github.ref, 'refs/tags/')
       run: |
         set -e
-        git for-each-ref github.ref --format='%(contents)' > release-notes.rst
+        git for-each-ref ${{ github.ref }} --format='%(contents)' > release-notes.rst
         # Strip PGP signature from signed tags
         sed -i "/-----BEGIN PGP SIGNATURE-----/,/-----END PGP SIGNATURE-----\n/d" release-notes.rst
     - name: Convert Release Notes to Markdown


webknjaz:
It'd be good to integrate zizmor...

seifertm:
@webknjaz Agreed!

I followed up with 2adcf52664e3e3b804505ed5ee09da83cc173fb8.

webknjaz:
@seifertm that still allows command injection just the same. The solution is to replace GH-level templating w/ Bash interpolation.

webknjaz:
I made a zizmor workflow recently. Here's how to integrate it: https://github.com/aio-libs/aiobotocore/pull/1355/files#diff-662dff05c7dfe6681c1007ae601e8573d413a03f9f9d53d951c2cb99caba4fd4R72

seifertm:
> @seifertm that still allows command injection just the same. The solution is to replace GH-level templating w/ Bash interpolation.

Right… Due to the Github template interpolation, you could insert `"` and escape the quoting. Thanks for pointing that out @webknjaz !

According to [the docs on default environment variables](https://docs.github.com/de/actions/writing-workflows/choosing-what-your-workflow-does/store-information-in-variables#default-environment-variables), we can simply use the `GITHUB_REF` env variable. This eliminates the Github template expression entirely: 0688d17581eb7396e5acfc14e8090d854f4e10aa

webknjaz:
Yep, for many simple things there's built-in env vars. For some, you might need to compute and put those into env vars. Note that in some contexts (like nested composite actions), some of the env vars may be borked, though.

PR #1122, opened 2025-05-15
Harden GitHub Actions workflow

seifertm:
This PR introduces [zizmor](https://docs.zizmor.sh/) for linting GitHub Actions Workflows and addresses all of its findings.

Zizmor is used both as a pre-commit hook and a separate job in the pipeline. This is due to the fact that some analyses require a GitHub token, which not all developers may have readily available.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -158,7 +158,9 @@ jobs:
       run: pip install .
     - name: Compile Release Notes Draft
       if: ${{ !contains(github.ref, 'refs/tags/') }}
-      run: towncrier build --draft --version "${{ needs.lint.outputs.version }}" > release-notes.rst
+      run: towncrier build --draft --version "${version}" > release-notes.rst
+      env:
+        version: ${{ needs.lint.outputs.version }}
     - name: Extract release notes from Git tag
       if: github.event_name == 'push' && contains(github.ref, 'refs/tags/')
       run: |
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -194,7 +194,7 @@ jobs:
       run: |
         tree dist
     - name: PyPI upload
-      uses: pypa/gh-action-pypi-publish@76f52bc884231f62b9a034ebfe128415bbaabdfc  # v1.12.4
+      uses: pypa/gh-action-pypi-publish@76f52bc884231f62b9a034ebfe128415bbaabdfc  # v1.12.4  # zizmor: ignore[use-trusted-publishing] # see #700
       with:
         attestations: true
         packages-dir: dist
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -102,7 +102,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - name: Decide whether the needed jobs succeeded or failed
-      uses: re-actors/alls-green@release/v1
+      uses: re-actors/alls-green@05ac9388f0aebcb5727afa17fcccfecd6f8ec5fe  # v1.2.2
       with:
         jobs: ${{ toJSON(needs) }}
     - uses: actions/checkout@v4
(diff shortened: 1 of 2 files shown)


.github/workflows/main.yml
```diff
@@ -95,17 +100,25 @@ jobs:
...
   check:
     name: Check
     if: always()
-    needs: [lint, test]
+    needs: [lint, lint-github-actions, test]
     runs-on: ubuntu-latest
     steps:
     - name: Decide whether the needed jobs succeeded or failed
-      uses: re-actors/alls-green@release/v1
+      uses: re-actors/alls-green@05ac9388f0aebcb5727afa17fcccfecd6f8ec5fe  # v1.2.2
```
webknjaz, line 116: It's possible to configure certain actions as trusted if you want.

.github/workflows/main.yml
```diff
@@ -95,17 +100,25 @@ jobs:
...
+    name: Lint GitHub Actions
+    permissions:
+      security-events: write
+    uses: zizmorcore/workflow/.github/workflows/reusable-zizmor.yml@1ae473d8672fe7613e809d86d202a35063736e16
+
   check:
     name: Check
     if: always()
-    needs: [lint, test]
+    needs: [lint, lint-github-actions, test]
```
webknjaz, line 112: Not sure if this is necessary — that workflow will always succeed. It reports the violations into the GH security UI instead. Do you think it makes sense to wait for it?

.github/workflows/main.yml
```diff
@@ -195,7 +213,7 @@ jobs:
         name: release-notes.md
         path: release-notes.md
     - name: GitHub Release
-      uses: ncipollo/release-action@v1
+      uses: ncipollo/release-action@440c8c1cb0ed28b9f43e4d1d670870f059653174  # v1.16.0
```
webknjaz, line 216: Not sure if it's wise to give this action OIDC privileges..

.github/workflows/main.yml
```diff
@@ -174,6 +190,8 @@ jobs:
     if: github.event_name == 'push' && contains(github.ref, 'refs/tags/')
     needs: [lint, check, prepare-release-notes]
     runs-on: ubuntu-latest
+    permissions:
+      id-token: write
```
webknjaz, line 194: If you're adding this, also remove password from the pypi-publish action..

.github/workflows/main.yml
```diff
@@ -95,17 +100,25 @@ jobs:
         path: coverage/coverage.*
         if-no-files-found: error
 
+  lint-github-actions:
+    name: Lint GitHub Actions
+    permissions:
+      security-events: write
+    uses: zizmorcore/workflow/.github/workflows/reusable-zizmor.yml@1ae473d8672fe7613e809d86d202a35063736e16
```
webknjaz, line 107:
```suggestion
    uses: zizmorcore/workflow/.github/workflows/reusable-zizmor.yml@3bb5e95068d0f44b6d2f3f7e91379bed1d2f96a8
```

.github/workflows/main.yml
seifertm: I understand your concern here. Do you see an alternative approach for creating GitHub releases as part of a package release? If we have similar without the action, we should consider removing it.

Other options:
- Rethink the release process entirely
- Review the pinned version of the action

seifertm:
> @seifertm looks like you marked my suggested change as resolved by mistake.

Indeed, thank you! I corrected that.

.github/workflows/main.yml
seifertm: That makes sense. I removed the job/workflow from the check step.

.github/workflows/main.yml
```diff
@@ -184,18 +202,17 @@ jobs:
       run: |
         tree dist
     - name: PyPI upload
-      uses: pypa/gh-action-pypi-publish@v1.12.4
+      uses: pypa/gh-action-pypi-publish@76f52bc884231f62b9a034ebfe128415bbaabdfc  # v1.12.4  # zizmor: ignore[use-trusted-publishing] # see #700
       with:
         attestations: true
         packages-dir: dist
```
webknjaz, line 208: I assume that the ignore was related to password that's no longer there? Also, all these inputs match defaults.
```suggestion
      uses: pypa/gh-action-pypi-publish@76f52bc884231f62b9a034ebfe128415bbaabdfc  # v1.12.4
```

PR #1141, opened 2025-06-10
test: Adapt unmarked async tests in strict mode for pytest 8.4.0

ziyao233:
Async tests fail instead of skipping and warning with Pytest 8.4.0 if no suitable async plugin is installed[1]. Adjust expectation of these tests to pass the testsuite with Pytest 8.4.0.

Link: https://docs.pytest.org/en/stable/changelog.html#pytest-8-4-0-2025-06-02 # [1]

### DIFF
--- a/tests/modes/test_strict_mode.py
+++ b/tests/modes/test_strict_mode.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 from textwrap import dedent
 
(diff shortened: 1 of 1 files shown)


seifertm:
That makes sense. Thanks for the contribution!

seifertm (approved):

Issue #1101, opened 2025-04-24
event_loop_policy for single test


Dreamsorcerer:
It would be good if there was an easy way to change the event_loop_policy for a single test. I can't see any way to do this currently from the documentation (other than moving a single test to it's own module).

In aiohttp, we currently have these fixtures in conftest.py, which I'm looking to remove:
```
@pytest.fixture
def selector_loop() -> Iterator[asyncio.AbstractEventLoop]:
    policy = asyncio.WindowsSelectorEventLoopPolicy()  # type: ignore[attr-defined]
    asyncio.set_event_loop_policy(policy)

    with loop_context(policy.new_event_loop) as:
        asyncio.set_event_loop(_loop)
        yield _loop


@pytest.fixture
def uvloop_loop() -> Iterator[asyncio.AbstractEventLoop]:
    policy = uvloop.EventLoopPolicy()
    asyncio.set_event_loop_policy(policy)

    with loop_context(policy.new_event_loop) as:
        asyncio.set_event_loop(_loop)
        yield _loop
```

seifertm:
That makes sense. The event loop policy support is not in a good state, unfortunately. I think this is closely related to #796.

seifertm:
I'm also linking #1032  which argues that this kind of customization should be possible through the _async_ marker.
For example: `@pytest.mark.asyncio(policy=…).

It's also worth mentioning that upstream is deprecating the policy system and looking to remove it entirely:
https://github.com/python/cpython/issues/127949

Dreamsorcerer:
> I think this is closely related to [#796](https://github.com/pytest-dev/pytest-asyncio/issues/796).

That one seems to just be concerned with sync tests geting executed multiple times.

> I'm also linking [#1032](https://github.com/pytest-dev/pytest-asyncio/issues/1032) which argues that this kind of customization should be possible through the _async_ marker.
> For example: `@pytest.mark.asyncio(policy=…)`.

Yes, this one may end up solving the issue.

Dreamsorcerer:
With a1, I'm now getting `pluggy._manager.PluginValidationError: unknown hook 'pytest_asyncio_loop_factories' in plugin <module 'conftest' from '/home/runner/work/aiohttp/aiohttp/tests/conftest.py'>`:
https://github.com/aio-libs/aiohttp/actions/runs/24638256636/job/72037513806

Dreamsorcerer:
Hmm, that might have only been happening on the free-threaded build..

[...]

Issue #1032, opened 2024-12-31
API for loop configuration options


asvetlov:
This issue doesn't propose a specific codebase change but is a point for discussing future steps in the plugin evolution.

Python has deprecated event loop policies since (unreleased) py-3.14.

The main asyncio motion idea for this and related changes is: 
1. Don't instantiate the loop explicitly, but use `Runner` (introduced in 3.11, there is a [backport](https://github.com/samypr100/backports.asyncio.runner) library for older Python version).
2. Control runner behavior by `debug` and `loop_factory` arguments.
3. There are eager tasks since 3.12; maybe the runner will have the option to configure it as well if we make an agreement

I think pytest-asyncio should reflect these changes.
1. `event_loop_policy()` fixture should be deprecated along with `event_loop`.
2. The library should not introduce public fixtures except `unused_tcp_port` and family. It is very fragile, especially if a user starts fixture overriding.
3. Fine-grained control could be done by special `pytest.mark.asyncio(...)` arguments.
4. Coarse-grained control uses configfile parameters and, maybe, cmdline args. 
5. Implementation details could vary; we are free to change them in the wild until the details are not leaked into the public space. The plugin already has a good movement in this direction, so I merely summarized what we do.

Please share your opinions in the comments, guys.

seifertm:
I'm not sure how we can make use of eager tasks, since I haven't used them, yet. My current understanding is that it is intended as a performance improvement.

Other than that, I totally agree with your points!

Thanks for pointing out the backport of asyncio.Runner! I wasn't aware of this, since there was no backport last time I checked. This could be the missing piece to properly tackle #127 and to simplify a lot of the code.

> I think pytest-asyncio should reflect these changes.
> 
>     1. `event_loop_policy()` fixture should be deprecated along with `event_loop`.
The only purpose of the _event_loop_policy_ is to allow testing against multiple event loops. The fixture was introduced due to lack of a better idea and is currently broken for code bases that mix sync and async tests (see #796). I already tried twice to fix this without success. The fact that the code base supports both the `event_loop` fixture and the newer `@pytest.mark.asyncio(loop_scope=…)` approach complicates things even more.

I'm in favor of deprecating it, but I think we should first provide a working replacement to parametrize tests over event loop implementations. I would be interested to see an API proposal how we want users to do this. Maybe something like `pytest.mark.asyncio(loop_factory=…)` and/or a config option? 

I fully agree with the rest of your points. Thanks for the push!

Issue #796, opened 2024-03-06
Parametrizing `event_loop_policy` parametrizes all tests


aceg1k:
Contrary to the statements in the documentation ([here](https://github.com/pytest-dev/pytest-asyncio/blob/b60649a648912a79ce6f3443ca7ac2c67fe9a5b7/docs/source/how-to-guides/multiple_loops.rst?plain=1#L5) and [here](https://github.com/pytest-dev/pytest-asyncio/blob/b60649a648912a79ce6f3443ca7ac2c67fe9a5b7/docs/source/reference/fixtures/index.rst?plain=1#L34-L35)), the fixture is not only applied to all pytest-asyncio tests, but also applied to all sync tests and thus sync tests are also run once for each parameter.

```python
import asyncio
import pytest
import uvloop

@pytest.fixture(
    scope="session",
    params=(asyncio.get_event_loop_policy(), uvloop.EventLoopPolicy()),
    ids=("asyncio", "uvloop"),
)
def event_loop_policy(request):
    return request.param

@pytest.mark.asyncio
async def test_async():
    pass

def test_sync():
    pass
```

Output:
```
plugins: asyncio-0.23.5
asyncio: mode=Mode.STRICT
collected 4 items                                                                                                       

test_event_loop_policy.py::test_async[asyncio] PASSED                                                             [ 25%]
test_event_loop_policy.py::test_sync[asyncio] PASSED                                                              [ 50%]
test_event_loop_policy.py::test_async[uvloop] PASSED                                                              [ 75%]
test_event_loop_policy.py::test_sync[uvloop] PASSED                                                               [100%]
```

seifertm:
Thanks for the report and the reproducer!

The event loop policy fixture is defined as an autouse fixture. I don't exactly recall why, but I think the reason is that it's now used by the deprecated `event_loop` fixture and had to be marked as autouse for backwards compatibility.

It's obviously a bug that this also parametrizes all sync tests and needs to be addressed.

However, I don't expect this issue to be fixed before v1.0 when the legacy `event_loop` fixture is gone.

sean-anderson-seco:
Ran into this today. Unfortunately, this means that `event_loop_policy` is useless for me. So the easiest way is to go back to `event_loop`:

```py
import asyncio
import contextlib
import warnings

import pytest
import uvloop

@pytest.fixture(
    scope="session",
    params=(asyncio.get_event_loop_policy(), uvloop.EventLoopPolicy()),
[...]

PR #1164, opened 2025-07-09
Make new_event_loop as a non-deprecated approch to running different eventloops

Vizonex:
After mentally thinking about this back and forth. I decided to step to the plate and try and fix this problem once and for all. 
The aiohttp developers have told me they had a bit of intrest utilizing this library and so would I with my plans to migrate from unittest to pytest for [winloop](https://github.com/Vizonex/winloop) and a new project coming that aims to upgrade & migrate [pycares to cython](https://github.com/Vizonex/cyares), which I've added my own deadline that I would move winloop to pytest by 2026. And yes [aiothreading](https://github.com/Vizonex/aiothreading) which aims to be an alternative to aiomultiprocess for users with slower hardware or providing safer solutions for heavier loads.

While talking with the aiohttp devs on the matrix about the problem. I decided to see if I could have a crack at providing a way to run eventloops without needing to use deprecated setups. Unfortunately I'm extremely picky about deprecated setups This might be a good thing however because the sooner someone does something to solve the problem the better.

### DIFF
--- a/pytest_asyncio/plugin.py
+++ b/pytest_asyncio/plugin.py
@@ -546,6 +546,21 @@ def _temporary_event_loop_policy(policy: AbstractEventLoopPolicy) -> Iterator[No
         _set_event_loop(old_loop)
 
 
+@contextlib.contextmanager
+def _temporary_event_loop(loop:AbstractEventLoop):
+    try:
+        old_event_loop = asyncio.get_event_loop()
+    except RuntimeError:
+        old_event_loop = None
+    
+    asyncio.set_event_loop(old_event_loop)
+    try:
+        yield 
+    finally:
+        asyncio.set_event_loop(old_event_loop)
+
+
+
 def _get_event_loop_policy() -> AbstractEventLoopPolicy:
     with warnings.catch_warnings():
         warnings.simplefilter("ignore", DeprecationWarning)
@@ -772,6 +787,9 @@ def _create_scoped_runner_fixture(scope: _ScopeName) -> Callable:
                             RuntimeWarning,
                         )
 
+    
+
+
     return _scoped_runner
 
 
@@ -779,6 +797,11 @@ for scope in Scope:
     globals()[f"_{scope.value}_scoped_runner"] = _create_scoped_runner_fixture(
         scope.value
     )
+
+@pytest.fixture(scope="session", autouse=True)
+def new_event_loop() -> AbstractEventLoop:
+    """Creates a new eventloop for different tests being ran"""
+    return asyncio.new_event_loop()
 
 
 @pytest.fixture(scope="session", autouse=True)


Vizonex:
> I agree that the existing approach with the `event_loop_policy` needs to be replaced (see also #1032). I'd love to deprecate it, but we first need another solution, as you said.
> 
> I'd like to avoid configuring the loop type through a fixture, though, for two reasons:
> 
> 1. A fixture invites users to add extra code that is unrelated to the configuration of the loop type. This makes it significantly harder for pytest-asyncio to ship non-breaking changes and is one of the main reasons the `event_loop` fixture got removed. I believe you mentioned in your other issue that you're familiar with the history, so I won't get into it.
> 2. The current `event_loop_policy` fixture takes the same approach as the `new_event_loop` fixture in this PR. The _autouse_ configuration causes non-async tests to be parametrized, leading to sync tests being executed twice (see [Parametrizing `event_loop_policy` parametrizes all tests #796](https://github.com/pytest-dev/pytest-asyncio/issues/796)). To my knowledge, there's currently [no way to apply autouse selectively.](https://github.com/pytest-dev/pytest/issues/1050#issuecomment-1257149605)
> 
> I'm currently leaning towards a configuration setting in _pyproject.toml_ dor a global parametrization of loop types. I don't really like it, so I'm open for alternative suggestions :) It definitely needs some more thought.
> 
> Besides parametrizing globally, we also want users to be able to override the loop type for specific tests (see #1032 #1101). Users should be able to pass a `loop_factory` keyword argument to the _asyncio_ marker. For example:
> 
> ```python
> @pytest.mark.asyncio(loop_factory=uvloop.new_event_loop)
> def test_run_with_uvloop():
>     ...
> ```
> 
> If you're interested in working on the `pytest.mark.asyncio(loop_factory=...)` support, I'll definitely support it!

That would be a great solution actually. Thanks for helping me come up with a compromise, I'll see what I can do to make that happen.

Vizonex:
Let me revert my changes and then I'll make new ones.

Vizonex:
Guess the only thing I will have to figure out is where loop-factory should get executed I wanted to make it so that both Sequences of factories and singular factories could be accepted.

Vizonex:
Now I think about it I am having trouble finding where to inject loop factories. I'll upload what I have so far and take a temporary break. But I'm trying to make it so that loop_factory can pass multiple factories and run them all.

Vizonex:
This puzzle turned out to be bigger than I thought let me push what I have now so far...

seifertm:
I had another look at the edge cases and how to address them. To be honest, I don't think it's reasonable to implement the proposed API.

The `loop_factory` argument to `pytest.mark.asyncio` and `pytest_asyncio.fixture` tries to configure the underlying `asyncio.Runner`. However, the runner is bound to a pytest scope, whereas the *asyncio* marker is bound to a test. Since there can be multiple tests within a single scope, pytest-asyncio would have to keep track of the current Runner's loop factory and make sure it doesn't get changed until the end of the scope.

The problem is exacerbated by the fact that tests may pull in fixtures, which can have their own *loop_factory* arguments and each fixture may depend on other fixtures. All in all, I think the amount of complexity isn't worth it.

I suggest we proceed with a simpler type of parametrization, for example specifying a *single* set of loop factories that applies to *all* tests.

PR #1172, opened 2025-07-24
Add pyright hook

### LINKED ISSUE #731: Decorator type signatures not recognized by pyright
...and [according to pyright upstream](https://github.com/microsoft/pyright/issues/3497#issuecomment-1134850832), this is a bug in mypy, rather than in pyright.

Type-checking pytest-asyncio itself yields a great many errors, starting with:

```none
pytest-asyncio/pytest_asyncio/plugin.py:57:60 - error: Type variable "_R" has no meaning in this context (reportGeneralTypeIssues)
pytest-asyncio/pytest_asyncio/plugin.py:57:36 - error: TypeVar bound type cannot be generic
pytest-asyncio/pytest_asyncio/plugin.py:60:65 - error: Type variable "_R" has no meaning in this context (reportGeneralTypeIssues)
pytest-asyncio/pytest_asyncio/plugin.py:60:37 - error: TypeVar bound type cannot be generic
```

...and type-checking a project that uses pytest-asyncio's decorators with `reportUntypedFunctionDecorator` set in pyright's configuration:

```
error: Untyped function decorator obscures type of function; ignoring decorator (reportUntypedFunctionDecorator)
```

---

I suspect that the option of _either_ passing the function to be used as a fixture as a direct argument to `fixture()` _or_ returning a function that acts as a proper decorator adds complexity here. If it's not possible to be pyright-friendly with the existing calling convention flexibility, might it make sense to offer a less-flexible, more explicitly-typed alternative?

tjkuson:
Hello, I was working with this plugin and encountered this issue.

I added pyright as a pre-commit check with exclusions matching the mypy hook.

This reveals one error involving `function.parent` possibly being `None`. I resolved this by adding an assertion, as that appears to be most consistent with the rest of the codebase.

The pre-commit hooks and test suite are passing locally.

Closes #731

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -78,6 +78,13 @@ repos:
   rev: v1.11.0
   hooks:
   - id: zizmor
+- repo: https://github.com/RobertCraigie/pyright-python
+  rev: v1.1.403
+  hooks:
+  - id: pyright
+    exclude: ^(docs|tests)/.*
+    additional_dependencies:
+    - pytest
 ci:
   skip:
   - actionlint-docker
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -78,14 +78,6 @@ repos:
   rev: v1.11.0
   hooks:
   - id: zizmor
-- repo: https://github.com/RobertCraigie/pyright-python
-  rev: v1.1.403
-  hooks:
-  - id: pyright
-    exclude: ^(docs|tests)/.*
-    additional_dependencies:
-    - pytest
-    - nodeenv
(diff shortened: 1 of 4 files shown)


seifertm:
Rather than running pyright in pre-commit, we can consider running it in a dedicated tox environment. This would give us the benefit of having a pyright check for the code base, while 90% of typing issues should still be caught by the mypy pre-commit hook.

Do you think this approach would side-step the issue?

seifertm (approved):
Thanks! This will help catch issues that are flagged by pyright, but not by mypy (even though we don't run pyright in a pre-commit hook).

PR #1173, opened 2025-07-25
Add debug option

### LINKED ISSUE #980: Enable "asyncio debug mode" for default event loop
The Python `asyncio` library has a debug mode, which is supposed to ease developing: https://docs.python.org/3/library/asyncio-dev.html#debug-mode

What about enabling this in `pytest-asyncio` for the default event loop? This could be made configurable via a config option.

tjkuson:
Add a config option and CLI option to use asyncio debug mode.

Closes #980


seifertm (approved):
No comments, great job. Thanks for the contribution!

PR #1180, opened 2025-08-09
Fix error when a shared event loop is unset

### LINKED ISSUE #1177: `RuntimeError` when event loop is unset between tests
When using a shared event loop (e.g., when `loop_scope="session"`) and `asyncio.set_event_loop(None)` is called from a test, any subsequent async test raises `RuntimeError: There is no current event loop in thread 'MainThread'`. 

This will happen any test that calls `asyncio.Runner` or `asyncio.run`, which call `asyncio.set_event_loop(None)` as part of its clean-up code.

To reproduce:

```python
import asyncio
import pytest


@pytest.mark.asyncio
async def test_before() -> None:
    pass


def test_set_event_loop_none() -> None:
    asyncio.set_event_loop(None)


@pytest.mark.asyncio
async def test_after() -> None:
    pass
```

```console
$ pytest -o 'asyncio_default_test_loop_scope=module' test_mre.py
```

The expected behaviour is for all tests to pass, but an exception is raised instead. The exception is not raised if either `test_before` or `test_after` are skipped.

I found issue #658 which is related; however, that issue is closed and related to fixtures (whereas this does not), so decided to create a new one instead.

I have a patch that resolves the issue by reinstating the event loop if needed. Let me know if a PR would be welcome!

tjkuson:
Automatically reinstates the event loop on the main thread when needed.

The coverage report suggests the patch is uncovered, but I think this is due to limitations of measuring coverage across processes.

Fixes #1177

Note: this is a work-in-progress proposal based on my understanding of the issue. If the maintainers prefer a different approach or consider the issue invalid, I’m happy to adjust or close this PR.

### DIFF
--- a/changelog.d/1172.added.rst
+++ b/changelog.d/1172.added.rst
@@ -1,0 +1,1 @@
+Compatibility with the `Pyright` type checker


pytest_asyncio/plugin.py
```diff
@@ -663,7 +670,12 @@ def wrap_in_sync(
     @functools.wraps(func)
     def inner(*args, **kwargs):
         coro = func(*args, **kwargs)
-        _loop = _get_event_loop_no_warn()
+        try:
+            _loop = _get_event_loop_no_warn()
+        except RuntimeError:
+            # Handle situation where asyncio.set_event_loop(None) removes shared loops.
+            _reinstate_event_loop_on_main_thread()
```
seifertm, line 677: The function `_get_event_loop_no_warn` is a helper that suppresses a call to `get_event_loop`, which is effectively deprecated in favor of `get_running_loop`. This deprecation was one of the drivers for removing the `event_loop` fixture, which relied on this functionality.

The fact that `wrap_in_sync` still uses this is unfortunate and I hope to rectify this in #1203.

The point I'm trying to make is that *wrap_in_sync* seems to be the wrong place to reinstate the loop, because it would mean doubling down on the *get_event_loop* use.

pytest_asyncio/plugin.py
```diff
@@ -602,6 +603,12 @@ def _set_event_loop(loop: AbstractEventLoop | None) -> None:
         asyncio.set_event_loop(loop)
 
 
+def _reinstate_event_loop_on_main_thread() -> None:
+    if threading.current_thread() is threading.main_thread():
```
seifertm, line 607: My understanding is that pytest doesn't support multi-threading. It doesn't hurt to have this check, but I wonder if you had a specific use case in mind that I'm missing. Is there? :)

seifertm (commented):
@tjkuson In the issue related to this PR, you [mentioned](https://github.com/pytest-dev/pytest-asyncio/issues/1177#issuecomment-3194472176) you had a local patch that satisfies the [edge case](https://github.com/pytest-dev/pytest-asyncio/issues/1177#issuecomment-3182183029) I mentioned. It sounded really promising and I'd appreciate if you could share it!

pytest_asyncio/plugin.py
tjkuson: I've removed the check. It was only there as it was needed in the helper function I used in the code under test to temporarily resolve this issue (which did happen to use multi-threading).

tjkuson:
@seifertm I've pushed the patch that satisfies the edge-case you mentioned.

pytest_asyncio/plugin.py
```diff
@@ -49,6 +49,16 @@
     PytestPluginManager,
 )
 
+_seen_markers: set[int] = set()
```
seifertm, line 52: I assume this code is intended to suppress duplicate warnings about keyword arguments in the *asyncio* marker. This should no longer be needed, because it has been addressed in https://github.com/pytest-dev/pytest-asyncio/pull/1203/commits/8585040230a04c512053db9c64d4410121a4ff58.

pytest_asyncio/plugin.py
```diff
@@ -602,6 +616,62 @@ def _set_event_loop(loop: AbstractEventLoop | None) -> None:
...
+    )
+)
+_function_loop: contextvars.ContextVar[asyncio.AbstractEventLoop | None] = (
+    contextvars.ContextVar(
+        "_function_loop",
+        default=None,
+    )
+)
+
+_SCOPE_TO_CONTEXTVAR = {
```
seifertm, line 650: The current way of handling contextvars with copying them and restoring them was introduced before switching to *asyncio.Runner*, if I remember correctly.

My understanding is that *asyncio.Runner* holds a default context along with its loop. We already have one Runner per loop scope. Do you think there's any way we can utilize the Runner's default context rather than storing contexts per scope externally?

This is just something to think about. If it's not possible or it would take a lot more work, we can start with this solution as well.

PR #1189, opened 2025-08-13
Fix: Add validation for `asyncio_default_fixture_loop_scope`

pankaj-bind:
This PR addresses an issue where the `asyncio_default_fixture_loop_scope` configuration option in `pytest.ini` was not being validated. This could lead to misconfigurations and unexpected behavior without providing any feedback to the user.

This change introduces a validation check in `pytest_asyncio/plugin.py` to ensure that the provided scope is one of the valid options (`function`, `class`, `module`, `package`, or `session`). If an invalid scope is provided, a `pytest.UsageError` is raised with a descriptive error message.

Additionally, a new test case has been added to `tests/test_fixture_loop_scopes.py` to verify that an invalid scope configuration correctly raises an error.

### DIFF
--- a/changelog.d/1189.added.rst
+++ b/changelog.d/1189.added.rst
@@ -1,0 +1,1 @@
+A ``pytest.UsageError`` for invalid configuration values of ``asyncio_default_fixture_loop_scope`` and ``asyncio_default_test_loop_scope``.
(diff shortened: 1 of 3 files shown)


seifertm (approved):
Thanks @pankaj-bind !

Issue #1191, opened 2025-08-16
Creating `anyio.CancelScope` in async fixture raises `Attempted to exit cancel scope in a different task than it was entered in`


provinzkraut:
```python
import anyio
import pytest


@pytest.fixture
async def scope():
    with anyio.CancelScope() as scope:
        yield scope


async def test(scope: anyio.CancelScope):
    pass
``` 

This will raise a `RuntimeError: Attempted to exit cancel scope in a different task than it was entered in`.

After looking at the code for fixture cleanup, I don't see an easy way to work around this, but it's quite a severe issue as it hinders working with cancel scopes / code that uses cancel scopes.

seifertm:
Based on the error message, I assume you use pytest-asyncio in *auto* mode and the `test` function is executed using pytest-asyncio. At least I can reproduce your error with pytest-asyncio v1.1.0 and anyio v4.10.0 when running `pytest --asyncio-mode=auto`.

Can you explain why your use case makes it necessary to use `anyio.CancelScope`, but run the test suite with pytest-asyncio as opposed to running the tests with anyio?

provinzkraut:
> At least I can reproduce your error with pytest-asyncio v1.1.0 and anyio v4.10.0 when running `pytest --asyncio-mode=auto`.

The code with explicit usage of `pytest-asyncio` and `--asyncio-mode=strict` gives the same result.

> Can you explain why your use case makes it necessary to use `anyio.CancelScope`, but run the test suite with pytest-asyncio as opposed to running the tests with anyio?

I'm working on the testing utilities of the [Litestar](https://github.com/litestar-org/litestar/) framework, which among other things provides to its users an async test client they may use in their test suites. This test client makes use of `anyio` and in particular `anyio.CancelScope`.
 
While it might be possible for Litestar's own test suite to drop `pytest-asyncio` in favour of `anyio`, this would also mean that all its downstream users would have to do the same, which does not sound great; A user should be able to use whatever async test runner they want.

seifertm:
[...]

PR #1193, opened 2025-08-17
Ensure async fixture setup and teardown run in the same task

provinzkraut:
An attempt to fix https://github.com/pytest-dev/pytest-asyncio/issues/1191.

Inspired by anyio's pytest plugin: https://github.com/agronholm/anyio/blob/b8e91a5cfe35c3b28f91ea8251674a871a7f4e26/src/anyio/_backends/_asyncio.py#L2222-L2239

### DIFF
--- a/tests/async_fixtures/test_async_fixtures.py
+++ b/tests/async_fixtures/test_async_fixtures.py
@@ -37,3 +37,14 @@ class TestAsyncFixtureMethod:
     @pytest.mark.asyncio
     async def test_async_fixture_method(self):
         assert self.is_same_instance
+
+
+@pytest.fixture()
+async def setup_and_teardown_tasks():
+    task = asyncio.current_task()
+    yield
+    assert task is asyncio.current_task()
+
+
+async def test_setup_and_teardown_tasks(setup_and_teardown_tasks):
+    pass
--- a/pytest_asyncio/plugin.py
+++ b/pytest_asyncio/plugin.py
@@ -16,6 +16,7 @@ from asyncio import AbstractEventLoop, AbstractEventLoopPolicy
 from collections.abc import (
     AsyncIterator,
     Awaitable,
+    Coroutine as CoroutineT,
     Generator,
     Iterable,
     Iterator,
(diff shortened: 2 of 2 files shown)


seifertm (commented):
Thanks for sharing this! In general, I think we can try something like this.

It expect this to be a very intrusive change that has the potential to break a lot. Therefore, the queue-based task runner needs to be behind a feature toggle, such as a configuration value and/or a command line option.

I'm not sure what the best approach is codewise. My very early guess it that it would help to create some kind of abstraction on top of asyncio.Runner, and move some existing log, such as handling of contextvars, there. In the next step, we can create our new queue-based implementation and eventually we can use the feature toggle to switch between them. This is just an idea, though. I have no clue how it plays out.

tests/async_fixtures/test_async_fixtures.py
```diff
@@ -37,3 +37,14 @@ async def async_fixture_method(self):
     @pytest.mark.asyncio
     async def test_async_fixture_method(self):
         assert self.is_same_instance
+
+
+@pytest.fixture()
```
seifertm, line 42: The test file you found uses the "old" way of writing tests directly in a module. In the past, these kinds of tests caused side effects (e.g. failing to clean up resources resulting in a ResourceWarning) that only appeared much later in the test suite. Therefore, we moved to using [Pytester](https://docs.pytest.org/en/stable/how-to/writing_plugins.html#testing-plugins) for all pytest-asyncio tests.

I'd appreciate, if you rewrote the test to use *Pytester,* even though it's mostly boilerplate. You can check the the file `test_function_scope.py` for an example.

Although it's not strictly necessary, I suggest to use strict mode in those tests (or both, if there's different behavior depending on the mode).

seifertm:
> Would you be okay with taking an iterative approach to this?
> 
> A first iteration could be basically what I've got here, with a feature toggle added.

That makes sense. Code that lies around in a branch doesn't help anyone. The change you're proposing would not just benefit Litestar, but also other users. If we manage to put it behind a feature flag and all existing tests are green, it should be fine.

The feature flag documentation needs to communicate clearly, that this is an experimental feature that may bring breaking changes without notice. That is, until we stabilize it.
Ideally, we will eventually be able switch to the queue-based approach, but I don't fully understand the consequences, yet.

provinzkraut:
> > Would you be okay with taking an iterative approach to this?
> > A first iteration could be basically what I've got here, with a feature toggle added.
> 
> That makes sense. Code that lies around in a branch doesn't help anyone. The change you're proposing would not just benefit Litestar, but also other users. If we manage to put it behind a feature flag and all existing tests are green, it should be fine.
> 
> The feature flag documentation needs to communicate clearly, that this is an experimental feature that may bring breaking changes without notice. That is, until we stabilize it. Ideally, we will eventually be able switch to the queue-based approach, but I don't fully understand the consequences, yet.

Sounds like a good approach! I probably won't have much time in the next 2 weeks, but I'm planning to pick this back up after!

PR #1198, opened 2025-08-19
fix typos flake8-asyncio -> pytest-asyncio

jakkdl:
Was reading through the changelog and found this amusing brain fart I introduced in #979
Turns out I've contributed to way too many flake8-* projects x)


jakkdl:
Haha no yeah flake8-async (without the "io") is what caused me to confuse the two, but it'll never have any checks about pytest-asyncio fixture misuse 😅

Looks great!

seifertm (approved):

PR #1274, opened 2025-10-28
Add free threaded Python 3.14t to the testing

cclauss:


cclauss:
I am unsure.  When I see benchmarks at the bottom of https://py-free-threading.github.io/examples/asyncio I expect that there will be a lot of exploration of using asyncio and free threading together.  If this is not needed then we can close.

seifertm:
Let's merge it, it's a good idea. I was just interested in your perspective.

PR #1298, opened 2025-11-23
trivial: Formatting in default fixture loop scope message

JayBazuzi:
This message is long and a bit difficult to read; a little formatting can make it easier to read.


webknjaz:
@JayBazuzi looks like you'll have to re-flow the paragraph now.

PR #1348, opened 2026-01-30
Promote warning to error when applying @pytest.fixture to async fixtures

### LINKED ISSUE #1202: Turn deprecation of applying `@pytest.fixture` to async fixtures in *strict* mode into an error
Pytest-asyncio emits a deprecation warning when an async fixture is decorated with `@pytest.fixture` in strict mode as v0.25 (see #979).

The goal of this issue is to follow through with the deprecation and convert the warning into an error.

This is a breaking change that involves a major version bump to pytest-asyncio.

Please note that the the same behavior has been [deprecated in pytest v8.4.0](https://docs.pytest.org/en/stable/changelog.html#pytest-8-4-0-2025-06-02) and will be removed in pytest v9. This means there's a connection to the minimum supported pytest version.

dzhalaevd:
### What was wrong?
using `@pytest.fixture` on async fixtures emitted only a deprecation warning, but plan was to make it an exception

Closes: #1202 
Related issue: #1202 

### How it was fixed?
- Raises ``pytest.UsageError`` instead of ``PytestDeprecationWarning`` when using ``@pytest.fixture`` on async fixtures in strict mode (I'm not really sure about ``pytest.UsageError`` because some different func uses usual ``ValueError``)
- Update test cases
- Add changelog entry

### DIFF
--- a/changelog.d/1202.breaking.rst
+++ b/changelog.d/1202.breaking.rst
@@ -1,0 +1,1 @@
+**Breaking change**: using ``@pytest.fixture`` on async fixtures in strict mode raises ``pytest.UsageError`` instead of ``PytestDeprecationWarning``
--- a/changelog.d/1202.breaking.rst
+++ b/changelog.d/1202.breaking.rst
@@ -1,1 +1,0 @@
-**Breaking change**: using ``@pytest.fixture`` on async fixtures in strict mode raises ``pytest.UsageError`` instead of ``PytestDeprecationWarning``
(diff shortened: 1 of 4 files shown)


seifertm:
Thanks for the initiative @dzhalaevd ! This is definitely valuable. 

Since this is a breaking change, it would require a major version bump to pytest-asyncio. I would like to coordinate this with other breaking changes, so I will not merge the PR quite yet.

PR #1379, opened 2026-03-16
fix: "The test […] is not an async function" when `asyncio` marker is

### LINKED ISSUE #810: "The test […] is not an async function" when `asyncio` marker is added via `pytest_collection_modifyitems()`
This one I've noticed while testing IPython. The simplest reproducer is:

`conftest.py`:

```python
import inspect


def pytest_collection_modifyitems(items):
    for item in items:
        if inspect.iscoroutinefunction(item.obj):
            item.add_marker("asyncio")
```

`test_foo.py`:

```python
async def test_foo():
    pass
```

With pytest-asyncio 0.23.6, I'm getting:

```pytb
========================================================= test session starts =========================================================
platform linux -- Python 3.11.8, pytest-8.1.1, pluggy-1.4.0
rootdir: /tmp/repro
plugins: asyncio-0.23.6
asyncio: mode=Mode.STRICT
collected 1 item                                                                                                                      

test_foo.py s                                                                                                                   [100%]

========================================================== warnings summary ===========================================================
test_foo.py::test_foo
  test_foo.py:5: PytestWarning: The test <Function test_foo> is marked with '@pytest.mark.asyncio' but it is not an async function. Please remove the asyncio mark. If the test is not marked explicitly, check for global marks applied via 'pytestmark'.
    async def test_foo():

test_foo.py::test_foo
  /tmp/repro/.venv/lib/python3.11/site-packages/_pytest/python.py:184: PytestUnhandledCoroutineWarning: async def functions are not natively supported and have been skipped.
  You need to install a suitable plugin for your async framework, for example:
    - anyio
    - pytest-asyncio
    - pytest-tornasync
    - pytest-trio
    - pytest-twisted
    warnings.warn(PytestUnhandledCoroutineWarning(msg.format(nodeid)))

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
=================================================== 1 skipped, 2 warnings in 0.01s ====================================================
```

With pytest-asyncio-0.21.1, the test is correctly run as marked.

JiwaniZakir:
Fixes #810

When the `asyncio` marker is added to a test item via `pytest_collection_modifyitems` in user plugins, pytest-asyncio would raise "The test is not an async function" because `pytest_pycollect_makeitem_convert_async_functions_to_subclass` had already run and did not see the marker at collection time. The root cause was that async `Function` items marked after collection were never converted to `PytestAsyncioFunction` subclasses. Adds a `pytest_collection_modifyitems` hook in `pytest_asyncio/plugin.py` with `trylast=True` that iterates collected items and calls `PytestAsyncioFunction.item_subclass_for` and `_from_function` to perform the conversion for any `Function` carrying an `asyncio` marker that was not already specialized. Verified by the existing test suite covering the `pytest_collection_modifyitems` marker injection scenario introduced alongside this fix.

### DIFF
--- a/pytest_asyncio/plugin.py
+++ b/pytest_asyncio/plugin.py
@@ -611,6 +611,19 @@ def pytest_pycollect_makeitem_convert_async_functions_to_subclass(
     hook_result.force_result(updated_node_collection)
 
 
+@pytest.hookimpl(trylast=True)
+def pytest_collection_modifyitems(items: list[Item]) -> None:
+    for i, item in enumerate(items):
+        if (
+            isinstance(item, Function)
+            and not isinstance(item, PytestAsyncioFunction)
+            and item.get_closest_marker("asyncio")
+        ):
+            specialized_item_class = PytestAsyncioFunction.item_subclass_for(item)
+            if specialized_item_class:
+                items[i] = specialized_item_class._from_function(item)
+
+
 @contextlib.contextmanager
 def _temporary_event_loop_policy(policy: AbstractEventLoopPolicy) -> Iterator[None]:
     old_loop_policy = _get_event_loop_policy()
--- a/tests/test_asyncio_mark.py
+++ b/tests/test_asyncio_mark.py
@@ -186,3 +186,33 @@ def test_asyncio_marker_uses_marker_loop_scope_even_if_config_is_set(
 
     result = pytester.runpytest("--asyncio-mode=auto")
     result.assert_outcomes(passed=1)
+
+
+def test_asyncio_mark_added_via_collection_modifyitems_is_recognized(
+    pytester: Pytester,
+):
+    """Regression test for #810.
+
+    When a user plugin adds the asyncio marker via pytest_collection_modifyitems,
+    pytest-asyncio should recognize the async function and run it correctly instead
+    of raising "The test is not an async function".
+    """
+    pytester.makeini("[pytest]\nasyncio_default_fixture_loop_scope = function")
+    pytester.makeconftest(
+        dedent("""\
+        import inspect
+
+        def pytest_collection_modifyitems(items):
+            for item in items:
+                if inspect.iscoroutinefunction(getattr(item, "obj", None)):
+                    item.add_marker("asyncio")
+        """)
+    )
+    pytester.makepyfile(
+        dedent("""\
+        async def test_foo():
+            pass
+        """)
+    )
+    result = pytester.runpytest("--asyncio-mode=strict")
+    result.assert_outcomes(passed=1)


JiwaniZakir:
Fair point -- I'll add a regression test that demonstrates the fix before updating the implementation.

golikovichev (commented):
Took this for a spin against #810. On `main` the reporter's reproducer fails as described, and with this branch it passes.

I also tried two cases beyond the bare reproducer, since the thread raised a doubt about whether a late-added marker still interacts correctly with parametrization and fixtures. Same `conftest.py` adding the marker in `pytest_collection_modifyitems`, plus:

```python
import pytest
import pytest_asyncio

@pytest_asyncio.fixture
async def async_fix():
    return 42

async def test_async_fixture(async_fix):      # consumes an async fixture
    assert async_fix == 42

@pytest.mark.parametrize("n", [1, 2, 3])       # parametrized
async def test_param(n):
    assert n > 0
```

On `main` all of these fail with the "not an async function" warning and the skip. On this branch all four pass, so the conversion in `pytest_collection_modifyitems` wires up the async fixture and the parametrized variants rather than only silencing the warning. `tests/test_asyncio_mark.py` stays green here too.

One suggestion: the regression test covers the no-fixture, no-parametrize case only. Given the concern raised earlier in the thread, it might be worth adding a case where the late-marked async test consumes an async fixture and/or is parametrized, so the wider behavior is locked in against future refactors. Glad to share the snippet I used if useful.

tjkuson:
The issue I saw before was with `pytest.mark.asyncio`-driven parametrization (e.g., loop factories, loop scope). If the marker is applied after pytest has parametrized the test items, then that's a problem not solved by the PR. I'd be curious how other pytest plugins resolve this (I'd imagine it involves some sort of hack).

golikovichev:
Following up on the parametrization concern. I tried to reproduce the `loop_scope` case on this branch (pytest 9.0.3, pytest-asyncio 1.3.1.dev84).

A conftest adds the marker in `pytest_collection_modifyitems` (so after `pytest_generate_tests`) with `loop_scope="module"`:

```python
def pytest_collection_modifyitems(config, items):
    for item in items:
        func = getattr(item, "function", None)
        if func is not None and inspect.iscoroutinefunction(func):
            if not any(m.name == "asyncio" for m in item.iter_markers()):
                item.add_marker(pytest.mark.asyncio(loop_scope="module"))
```

Two module-level async tests record `id(asyncio.get_running_loop())`. To keep the check honest: with a bare `pytest.mark.asyncio` (function scope) the two ids differ; with `loop_scope="module"` they match. Adding the marker late, the module scope is still honored - both tests share one loop, and the same holds when the test is also `@pytest.mark.parametrize`d (every variant lands on the module loop).

I couldn't get `loop_factory` to apply at all though, late or via the decorator - `mark.asyncio` rejects it with "accepts only a keyword argument 'loop_scope'", so that one looks like a separate concern.

So at least for `loop_scope` I couldn't reproduce a miss with the marker added after parametrization. If the case you hit was a different shape (a loop-factory plugin, or an older version), a minimal repro would help and I'm glad to dig further.

tjkuson:
That's because the branch is out of date with main and does not contain the loop factory feature. The marker drives parametrization, so we'll need a solution for late parametrization.

tjkuson:
I've spent some time on this problem this weekend and I don't see a way to resolve this issue completely without altering the class-based, collection-time setup and ownership of test items, which would be a small breaking change (a small one though, I think). I'll tidy up the proof of concept and post it as a draft PR within the next few days.

golikovichev:
Thanks, you're right - I was on the old branch, so loop_factory was not there to test. Pulled main and I can see it now: the loop_factory parametrization happens via `metafunc.parametrize(_asyncio_loop_factory, ...)` in `pytest_generate_tests`, driven by the marker.

That explains why a late marker can not drive it: `pytest_generate_tests` runs during collection, before `pytest_collection_modifyitems`, so by the time a conftest adds the marker the parametrize has already been skipped, and it can not be recreated after the fact. That lines up with your conclusion that this needs a change at collection-time rather than a post-collection patch.

Happy to review the draft PoC when you post it.

seifertm:
This PR hasn't seen a reacting to the review comments from the OP in a while. I'll close the PR for now, but I'm happy to reopen it, if the OP is willing to continue and engage with the issues raised during the review.

PR #1381, opened 2026-03-17
fix: Finish deprecation of unset asyncio_default_fixture_loop_scope

### LINKED ISSUE #924: Finish deprecation of unset asyncio_default_fixture_loop_scope

JiwaniZakir:
Fixes #924

Completes the deprecation of the unset `asyncio_default_fixture_loop_scope` configuration option. The deprecation warning introduced in a prior release was the interim step; the final behavior is to default the fixture loop scope to `"function"` rather than requiring users to set it explicitly. Changes `default=None` to `default="function"` in `pytest_addoption` and removes the `_DEFAULT_FIXTURE_LOOP_SCOPE_UNSET` warning string and its associated `warnings.warn` call from `pytest_configure` in `pytest_asyncio/plugin.py`. Verified by confirming no warning is emitted when `asyncio_default_fixture_loop_scope` is unset and that fixture loop scope behavior matches the `"function"` default.

### DIFF
--- a/pytest_asyncio/plugin.py
+++ b/pytest_asyncio/plugin.py
@@ -113,7 +113,7 @@ def pytest_addoption(parser: Parser, pluginmanager: PytestPluginManager) -> None
         "asyncio_default_fixture_loop_scope",
         type="string",
         help="default scope of the asyncio event loop used to execute async fixtures",
-        default=None,
+        default="function",
     )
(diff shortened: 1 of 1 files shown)


seifertm:
Closing this PR in favor of #1444 , which already went through a review.

Issue #1382, opened 2026-03-17
Remove dependency on ncipollo/release-action


seifertm:
The goal of this issue is to replace *[ncipollo/release-action](https://github.com/ncipollo/release-action)* in the GitHub workflow with a call to *gh release create*.

see also https://github.com/pytest-dev/pytest-asyncio/security/code-scanning/23

PR #1392, opened 2026-03-30
feat(workflows): add gh release instead of ncipollo/release-action

dzhalaevd:
### What was wrong?

Due to security issues with `ncipollo/release-action` need to remove this dependency and replace with `gh release`

Closes: #1382
Related: #1382

### How it was fixed?

The release logic previously implemented via `ncipollo/release-action` has been reimplemented using the `gh`

Our workflow now relied on two key flags:
- `allowUpdates: true`
- `skipIfReleaseExists: true`

These flags overlap in behavior at least semantically and the [ncipollo documentation](https://github.com/ncipollo/release-action)￼ doesn't clearly define their interaction or precedence. Based on the existing workflow, i was reimplemented as follows with this idea:

```sh
if release exists:
    if draft:
        gh release edit
    if published:
        exit
else:
    gh release create
```

I tried run jobs for check how release creating [here](https://github.com/dzhalaevd/pytest-asyncio/releases)
and re-run for checking how it was updates

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -195,25 +195,15 @@ jobs:
       if: github.event_name == 'push' && contains(github.ref, 'refs/tags/')
       env:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-        VERSION: ${{ needs.build.outputs.version }}
         PRERELEASE: ${{ needs.build.outputs.prerelease }}
       run: |
         TAG_NAME="${GITHUB_REF#refs/tags/}"
 
-        IS_DRAFT=$(gh release view "${TAG_NAME}" --json isDraft --jq '.isDraft' 2>/dev/null || echo "")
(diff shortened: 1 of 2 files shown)


.github/workflows/main.yml
```diff
@@ -193,16 +193,32 @@ jobs:
...
-        draft: true
-        skipIfReleaseExists: true
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        VERSION: ${{ needs.build.outputs.version }}
+        PRERELEASE: ${{ needs.build.outputs.prerelease }}
+      run: |
+        TAG_NAME="${GITHUB_REF#refs/tags/}"
+
+        IS_DRAFT=$(gh release view "${TAG_NAME}" --json isDraft --jq '.isDraft' 2>/dev/null || echo "")
```
seifertm:
```suggestion
        gh release view "${TAG_NAME}" &>/dev/null
```

Issue #641, opened 2023-10-23
Allow switching doc examples between auto and strict mode


seifertm:
Users of pytest-asyncio either use auto mode or strict mode. They cannot mix modes within the same project. Some code examples in the documentation are presented for each mode, but the user is only interested for the examples in one of the modes.

I propose to add [sphinx-tabs](https://pypi.org/project/sphinx-tabs/) to the doc dependencies and use [group tabs](https://sphinx-tabs.readthedocs.io/en/latest/#group-tabs) to show only code examples for the selected pytest-asyncio mode.

Code examples that are only available for one mode should be provided for both, auto mode and strict mode as part of this issue.

seifertm:
Sounds good, thank you!

Please do file the PR even if you don't manage to go through all examples in the docs. A partial documentation update with mode switching is better than none at all.

Let me know if you run into any issues.

> My updated plan:
> 
>     1. Add sphinx-tabs to docs/requirements.txt
> 

Small nit: it should go to the `optional-dependencies.docs` section in `pyproject.toml` and to `constraints.txt`.

PR #1395, opened 2026-04-05
feat(docs): add tabs and group tabs to improve documentation examples

dzhalaevd:
### What was wrong?
Users of pytest-asyncio can only pick `auto` or `strict` mode per project, but the docs show examples for both. Is kinda confusing, so we need add tabs

Closes: #641
Related: #641

### How it was fixed?
- Add sphinx-tabs deps
- Update some doc files


seifertm (approved):
Thanks!

PR #1421, opened 2026-04-19
Deprecate `event_loop_policy` fixture

### LINKED ISSUE #1419: Deprecate event_loop_policy fixture
The asyncio policy system is deprecated. The *event_loop_policy* fixture should be deprecated in favor of the *pytest_asyncio_loop_factories* hook. The deprecation is needed to inform users to migrate away from the fixture.

The goal of this task is that every user who reimplements *event_loop_policy* receives a *DeprecationWarning*. The warning message should state that the *event_loop_policy* fixture is deprecated and will be removed. It should also explain what functionality users should use instead (i.e. the *pytest_asyncio_loop_factories* hook).

## Possible implementation
A similar mechanism was needed for the deprecation of the *event_loop* fixture. This has previously been implemented by [setting an attribute](https://github.com/pytest-dev/pytest-asyncio/blob/b24de088dbb92de2cedc506bcc4adb522a72bf7c/pytest_asyncio/plugin.py#L1127-L1132) on the original fixture definition provided by pytest-asyncio. This attribute was used to [emit a warning during fixture setup](https://github.com/pytest-dev/pytest-asyncio/blob/b24de088dbb92de2cedc506bcc4adb522a72bf7c/pytest_asyncio/plugin.py#L852-L858). Consequently, every fixture definition that didn't set the attribute caused a warning to be emitted.

This is just one way of solving this and merely provided as a pointer

tjkuson:
Emit a deprecation warning when a user uses the `event_loop_policy` fixture (determined by the hook `__module__`). Add deprecation notice to documentation.

One documentation code snippet tests was failing due to the change. As I was about to update it, I noticed that the snippet wasn't actually used anywhere, so I deleted it.

Closes #1419


seifertm (approved):
I like the mechanism to check whether the fixture has been overridden and adding a deprecation notice to the docs, instead of removing it from the documentation. Both are improvents on how pytest-asyncio handled theses things in previous versions.

Thanks!

PR #1430, opened 2026-04-30
Fix unclosed event loop

### LINKED ISSUE #724: ResourceWarning: unclosed event loop
using pytest-asyncio asyncio 0.23.2

```python
import pytest

import pytest_asyncio


@pytest_asyncio.fixture()
async def demo():
    yield


@pytest.mark.asyncio
async def test_aiohttp_test_client_json(demo):
    pass


def test_redirect():

    import asyncio
    async def amain():
        pass
    asyncio.run(amain())

    import gc
    gc.collect()
```

results in

```
REQUESTS_CA_BUNDLE=/home/graingert/projects/vcrpy/.tox/py311-aiohttp/lib/python3.11/site-packages/pytest_httpbin/certs/cacert.pem ./.tox/py311-aiohttp/bin/pytest tests/integration/test_aiohttp.py -s
================================================================================ test session starts =================================================================================
platform linux -- Python 3.11.7+, pytest-7.4.3, pluggy-1.3.0
rootdir: /home/graingert/projects/vcrpy
configfile: pyproject.toml
plugins: aiohttp-1.0.5, httpbin-2.0.0, cov-4.1.0, asyncio-0.23.2
asyncio: mode=Mode.STRICT
collected 2 items                                                                                                                                                                    

tests/integration/test_aiohttp.py ..

================================================================================== warnings summary ==================================================================================
tests/integration/test_aiohttp.py::test_redirect
  /usr/lib/python3.11/asyncio/base_events.py:692: ResourceWarning: unclosed event loop <_UnixSelectorEventLoop running=False closed=False debug=False>
    _warn(f"unclosed event loop {self!r}", ResourceWarning, source=self)
  Enable tracemalloc to get traceback where the object was allocated.
  See https://docs.pytest.org/en/stable/how-to/capture-warnings.html#resource-warnings for more info.

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
============================================================================ 2 passed, 1 warning in 0.02s ===========================================================================
```

tjkuson:
pytest-asyncio temporarily switches the event loop policy whilst setting up managed tests and fixtures. Before this change, that policy switch also called `asyncio.get_event_loop()` to save the current loop and restore it later.

`asyncio.get_event_loop()` can create a loop when none is set, meaning pytest-asyncio recorded the newly created loop as if it were a pre-existing one, restored it after the runner finished, and did not close it. A later test calling `asyncio.run()` could clear the current loop, allowing that loop to be collected and reported as `ResourceWarning: unclosed event loop`.

This makes `_temporary_event_loop_policy` only swap and restore the policy. The scoped runner fixture still installs the runner loop explicitly for custom loop factories, where `asyncio.Runner(loop_factory=...)` does not do so itself, and clears it during teardown.

By returning `asyncio.get_event_loop()` to its original behaviour after an async test, it now means that any subsequent sync tests will see a closed event loop. I think this likely aligns with how users would expect pytest-asyncio to manage the event loop.

Closes #724

### DIFF
--- a/changelog.d/724.fixed.rst
+++ b/changelog.d/724.fixed.rst
@@ -1,0 +1,1 @@
+Fixed a ``ResourceWarning: unclosed event loop`` warning that could occur when a synchronous test called ``asyncio.run()`` or otherwise unset the current event loop after pytest-asyncio had run an async test or fixture.
--- a/pytest_asyncio/plugin.py
+++ b/pytest_asyncio/plugin.py
@@ -806,23 +806,13 @@ def _temporary_event_loop(loop: AbstractEventLoop) -> Iterator[None]:
 @contextlib.contextmanager
 def _temporary_event_loop_policy(
     policy: AbstractEventLoopPolicy,
-    *,
-    has_custom_factory: bool,
 ) -> Iterator[None]:
     old_loop_policy = _get_event_loop_policy()
-    if has_custom_factory:
-        old_loop = None
-    else:
-        try:
-            old_loop = _get_event_loop_no_warn()
-        except RuntimeError:
-            old_loop = None
     _set_event_loop_policy(policy)
     try:
         yield
     finally:
         _set_event_loop_policy(old_loop_policy)
-        _set_event_loop(old_loop)
 
 
 def _get_event_loop_policy() -> AbstractEventLoopPolicy:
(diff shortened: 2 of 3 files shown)


seifertm (approved):
Thanks for the detailed explanation. We should absolutely get rid of any remaining use of _get_event_loop_ .

PR #1431, opened 2026-05-01
Add pytest main CI smoke test

### LINKED ISSUE #1132: CI: add job that that uses pytest-dev version

Herrtian:
Fixes #1132.

This adds a separate non-blocking CI job that installs pytest from the pytest main branch and runs the existing test target against it. The regular tox matrix stays unchanged.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -101,7 +101,7 @@ jobs:
         if-no-files-found: error
 
   test-pytest-dev:
-    name: ubuntu - Python ${{ env.PYTHON_LATEST }} - pytest main
+    name: ubuntu - Python latest - pytest main
     runs-on: ubuntu-latest
     continue-on-error: true
     steps:
@@ -116,11 +116,9 @@ jobs:
         python -VV
         python -m site
         python -m pip install --upgrade pip
-        python -m pip install --upgrade \
-          ".[testing]" \
-          "pytest @ git+https://github.com/pytest-dev/pytest.git"
+        python -m pip install --upgrade tox
     - name: Run tests against pytest main
-      run: make test
+      run: python -m tox run -e pytest-dev
 
   lint-github-actions:
     name: Lint GitHub Actions
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -100,6 +100,28 @@ jobs:
         path: coverage/coverage.*
         if-no-files-found: error
 
+  test-pytest-dev:
+    name: ubuntu - Python ${{ env.PYTHON_LATEST }} - pytest main
+    runs-on: ubuntu-latest
+    continue-on-error: true
+    steps:
+    - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd
+      with:
+        persist-credentials: false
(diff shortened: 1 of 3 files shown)


Herrtian:
The failing timeline check is for the missing changelog fragment. Since this is only adding a CI smoke test and doesn't change runtime behavior or docs, I think this should probably use the `skip-changelog` label rather than adding a release note.

.github/workflows/main.yml
```diff
@@ -100,6 +100,28 @@ jobs:
         path: coverage/coverage.*
         if-no-files-found: error
 
+  test-pytest-dev:
+    name: ubuntu - Python ${{ env.PYTHON_LATEST }} - pytest main
```
seifertm: The [workflow run](https://github.com/pytest-dev/pytest-asyncio/actions/runs/25220012342) fails and complains about *env* not being found. I assume the reason is that *env* just [isn't available](
https://docs.github.com/en/actions/reference/workflows-and-actions/contexts#context-availability) here. We can just use "latest", instead of the variable value.

seifertm (changes requested):
Thanks for the initiative @Herrtian !

It's desirable to have parity between the CI/CD workflow and the local development environment. That means, it should be possible to run the tests against pytest dev in the local development environment.

There was another [Pull Request](https://github.com/pytest-dev/pytest-asyncio/pull/1380/changes) which seems like it had this figured out. Can you take a look and extend the local development environment so that there's a tox environment that runs all tests against the development version of pytest?

Herrtian:
Thanks, updated this in 09a6a37. The job name now uses a literal latest, and the CI job runs a new local tox env, pytest-dev, so the same pytest-main check can be run outside CI too. I checked the tox config locally and ran the env against pytest main. It currently surfaces a few strict-mode behavior failures from pytest main, so I kept the CI job non-blocking as originally intended.

seifertm (approved):
Test errors in the pytest-dev environment are related to pytest's stricter handling of async fixtures, which now gives an error, instead of a warning.

seifertm:
Thanks @Herrtian !

seifertm:
> Test errors in the pytest-dev environment are related to pytest's stricter handling of async fixtures, which now gives an error, instead of a warning.

Actually, no. This was introduced as early as v8.4.0. Something else is off here.

seifertm:
https://github.com/pytest-dev/pytest/pull/14015 is the responsible change. We need to adjust the tests accordingly.

seifertm:
I decided to skip the offending tests on Pytest 9.1 onwards. They will no longer be needed in pytest-asyncio once the minimum pytest version is bumped to v9.1, because the logic is already handled in pytest.

seifertm (approved):

PR #1432, opened 2026-05-04
Document selecting loop factories per test

tjkuson:
Document in more detail how to select loop factories per test (specifically, how the hook interacts with the `loop_factories` parameter).

The gist of what this is trying to do is document how `loop_factories` works alongside pytest's ordinary hook semantics.

Relates to https://github.com/pytest-dev/pytest-asyncio/issues/1101#issuecomment-4361757705

CI is failing as I didn't update the changelog. My understanding is that this PR can skip the changelog as the documentation hasn't been released yet.

### DIFF
--- a/docs/how-to-guides/run_test_with_specific_loop_factories.rst
+++ b/docs/how-to-guides/run_test_with_specific_loop_factories.rst
@@ -2,15 +2,22 @@
 How to run a test with specific event loop factories only
 =========================================================
 
-To run a test with only a subset of configured factories, use the ``loop_factories`` argument of ``pytest.mark.asyncio``:
+``pytest_asyncio_loop_factories`` determines which named event loop factories are available for each test item.
+By default, pytest-asyncio parametrizes a test with every factory returned for that item.
+Use ``loop_factories`` to select a subset of the factory names returned by the hook.
 
-.. code-block:: python
+Assume ``conftest.py`` provides two named factories:
 
-   import pytest
+.. include:: run_test_with_specific_loop_factories/conftest.py
+   :code: python
 
+Then use ``loop_factories`` to select which available factory names a test should run with:
 
-   @pytest.mark.asyncio(loop_factories=["custom"])
-   async def test_only_with_custom_event_loop():
-       pass
+.. include:: run_test_with_specific_loop_factories/test_loop_factories_subset.py
+   :code: python
 
 If a requested factory name is not available from the hook, the test variant for that factory is skipped.
+
+For declaring the factories themselves, see :doc:`custom_loop_factory`.
+
+For choosing the available factories from the pytest item, see :doc:`configure_loop_factories_per_test`.
--- a/docs/how-to-guides/configure_loop_factories_per_test.rst
+++ b/docs/how-to-guides/configure_loop_factories_per_test.rst
@@ -1,0 +1,29 @@
+========================================================
+How to configure event loop factories from the test item
+========================================================
+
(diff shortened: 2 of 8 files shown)


docs/how-to-guides/run_test_with_specific_loop_factories.rst
```diff
@@ -14,3 +16,45 @@ To run a test with only a subset of configured factories, use the ``loop_factori
        pass
 
 If a requested factory name is not available from the hook, the test variant for that factory is skipped.
+
+The hook receives the test item, so it can choose which factories are available for that test.
+For example, it can inspect markers, fixture names, or the test node ID:
+
+.. code-block:: python
```
seifertm: The pytest-asyncio test suite discovers tests in Python files under the *docs* directory. This allows executing the code examples in the documentation and make it less likely for them to be outdated or break. This is currently done via an *include* directive, such as in *parametrize_with_asyncio.{rst.py}*.

Would it be possible to use the same approach for these two code examples? I imagine it would just work when grouping both of them into a subfolder under *how-to-guides/*.

If it doesn't work straight away, we leave it as is.

docs/how-to-guides/run_test_with_specific_loop_factories.rst
seifertm, line 1: The first part up to line 18 explains that all tests are parametrized with whatever entries are returned by the hook and that users can limit the parametrization via the *loop_factories* argument in the asyncio marker. The added code examples show an "inverted" control flow: there's no parametrization by default, unless the user explicitly lists the factory identifier in *loop_factories*.

I understand that we want to show both and each content makes sense, but it could be confusing that two approaches are mixed.

I'm not sure how to best solve this. We could either highlight that these are two very different approaches, e.g. "Variant 1, Variant 2" headlines or keep both variants in separate how-tos. I'm leaning towards the latter, but I'm failing to come up with distinguishable how-to titles.

seifertm (changes requested):

docs/how-to-guides/run_test_with_specific_loop_factories.rst
tjkuson: Seems to work. 0a1803deb126fede7e804cf1e0e45acd0733328b

docs/how-to-guides/run_test_with_specific_loop_factories.rst
tjkuson: Yeah, I see what you mean.

I revisited how Diátaxis works and I think there's a reasonable argument in favour of this being three separate how-to guides: how to declare the event loop factories; how to declare factories based on item-specific logic; and how to select a subset of available factories from a test. I am curious what you think: 0a1803deb126fede7e804cf1e0e45acd0733328b f1a4d0d0853f19def52b2e578cbeaf9312e7fa50

seifertm:
Sorry for the delay on this… I promise to take a look tomorrow.

docs/how-to-guides/run_test_with_specific_loop_factories.rst
seifertm: The separation makes sense to me. I appreciate that these three how-tos reference each other. The difference is quite nuance and the cross-reference makes it easier for the user to find the right section in case they took a wrong turn in the docs.

seifertm (approved):
Squashed the commits before merging.

Thanks again for the help here!

tjkuson:
> Sorry for the delay on this… I promise to take a look tomorrow.

No need to apologise, I appreciate the work you do here!

PR #1444, opened 2026-05-12
Set default asyncio fixture loop scope to function

### LINKED ISSUE #924: Finish deprecation of unset asyncio_default_fixture_loop_scope

SahilArchitect:
Closes #924.

  ## Summary

  Sets the default value of `asyncio_default_fixture_loop_scope` to `function` and removes the deprecated unset
  fallback/warning path.

  This intentionally only changes the default for async fixture loop scope. It does not change async test loop scope
  behavior, custom `event_loop` fixture deprecations, pytest warning filtering order, or quoted `pytest.ini` value
  parsing.

  ## Changes

  - Default unset `asyncio_default_fixture_loop_scope` to `function`
  - Remove the old fallback to fixture caching scope
  - Update docs and changelog wording
  - Add regression coverage for the unset/default behavior

  ## Tests

  - `python -m pytest tests/test_fixture_loop_scopes.py`
  - `python -m pytest tests/test_asyncio_fixture.py tests/modes/test_auto_mode.py tests/modes/test_strict_mode.py`


changelog.d/1298.changed.rst
seifertm, line 1: The news fragment describes a change of a warning message. Your change is unrelated to this news fragment and will be included in a different release. Please remove any modifications to this file.

docs/how-to-guides/migrate_from_0_21.rst
seifertm, line 1: This is the migration guide for users looking to upgrade from an older version of pytest-asyncio. This patch doesn't change anything in the upgrade path. Please remove any modifications to this file.

docs/how-to-guides/migrate_from_0_23.rst
seifertm, line 1: This is the migration guide for users looking to upgrade from an older version of pytest-asyncio. This patch doesn't change anything in the upgrade path. Please remove any modifications to this file.

tests/test_fixture_loop_scopes.py
```diff
@@ -35,6 +35,29 @@ async def test_runs_in_same_loop_as_fixture(fixture):
     result.assert_outcomes(passed=1)
 
 
+def test_default_fixture_loop_scope_is_function_when_unset(pytester: Pytester):
```
seifertm, line 38: How does this test assert that the fixture is function-scoped? In other words, how would the test fail if the fixture was not function-scoped?

seifertm (changes requested):
Thanks for the initiative, @SahilArchitect .

This is a breaking change that requires a major version bump. The next version will be v1.4.0, so I won't merge the PR just yet. In the meantime, could you have a look at my comments?

SahilArchitect:
Thanks for the review. I pushed a follow-up commit now.

What I changed:
- restored `changelog.d/1298.changed.rst`, since that was for the older warning-message change
- restored both migration guide files, since this PR does not need to change those upgrade docs
- updated the regression test so it now actually proves the unset default is using `function`

For the test: I changed it to use a `scope="module"` async fixture with no explicit `loop_scope`. With the old behavior, that would fall back to the module fixture scope and pass. With the new behavior, it tries to use `_function_scoped_runner`, so we get the expected `ScopeMismatch`. That makes the behavior difference visible in the test.

I also reran:

```
.venv/bin/python -m pytest tests/test_fixture_loop_scopes.py -q
.venv/bin/python -m pytest tests/test_asyncio_fixture.py tests/modes/test_auto_mode.py tests/modes/test_strict_mode.py -q
```

Both passed.

tests/test_fixture_loop_scopes.py
seifertm: @SahilArchitect You resolved all other comments, but ignored this one :)
Now, the tests use module scope and have nothing to do with function scope anymore.

seifertm (changes requested):

tests/test_fixture_loop_scopes.py
SahilArchitect: You are right, I misunderstood this one. The previous version was proving a mismatch with module scope, not clearly proving the fixture loop is function-scoped.

I pushed another small commit and changed the test to use a normal async fixture without `loop_scope`. It now runs two async tests and checks two things:

- inside each test, the fixture loop is the same as the currently running test loop
- the second test gets a different fixture loop from the first one

So if the fixture loop was module/session scoped, this would fail instead of passing.

Sorry for missing the point the first time.

SahilArchitect:
You are right, I misunderstood this one. The previous version was proving a mismatch with module scope, not clearly proving the fixture loop is function-scoped.

I pushed another small commit and changed the test to use a normal async fixture without loop_scope. It now runs two async tests and checks two things:

inside each test, the fixture loop is the same as the currently running test loop
the second test gets a different fixture loop from the first one
So if the fixture loop was module/session scoped, this would fail instead of passing.

Sorry for missing the point the first time.

Issue #1346, opened 2026-01-26
Documented `uvloop` pattern is deprecated


tjkuson:
Currently, the [documentation suggests](https://github.com/pytest-dev/pytest-asyncio/blob/b817cf476a6dda2fb4a847cfd1b3285209598779/docs/how-to-guides/uvloop.rst?plain=1#L8)

```python
import pytest
import uvloop


@pytest.fixture(scope="session")
def event_loop_policy():
    return uvloop.EventLoopPolicy()
```

`EventLoopPolicy` was deprecated in https://github.com/MagicStack/uvloop/pull/638.

golikovichev:
+1 on ivanthewebber's note about the text/example mismatch. As a downstream user adopting pytest-asyncio for new projects, the current snippet is misleading on two fronts:

1. `uvloop.EventLoopPolicy()` is deprecated since uvloop 0.21 (per MagicStack/uvloop#638) - calling it in the fixture emits a `DeprecationWarning` even before the test runs.
2. Python 3.14 deprecates the policy system entirely per seifertm's note above, so any interim doc that still hangs off `event_loop_policy` will age fast.

Two minimal interim options for the docs page (not blocking the long-term #1164 work):

a) Replace the snippet with the `asyncio.Runner(loop_factory=uvloop.new_event_loop)` pattern (works today, sidesteps deprecation, but requires the user to wire it through pytest-asyncio).
b) Keep the existing snippet but add a one-line callout noting both deprecations and pointing at #1164 for the future API direction.

Happy to PR option (b) - one-paragraph callout, no API contract change - if maintainers prefer that as a holding pattern.

tjkuson:
For what it's worth, this should be resolved once 1.4.0 is released ([pre-releases are available](https://github.com/pytest-dev/pytest-asyncio/releases/tag/v1.4.0a2) if you'd like to try it out early).

golikovichev:
Thanks for the heads-up. Will try v1.4.0a2 to confirm the new pattern. Dropping the interim docs callout offer since the structural fix is on its way.

seifertm:
#1373 improves the documentation by mentioning the deprecation of the policy system and providing a non-deprecated approach in the how-to. I'd say this issue is sufficiently addressed.

PR #1454, opened 2026-05-21
Fix event_loop_policy parametrization scope

### LINKED ISSUE #796: Parametrizing `event_loop_policy` parametrizes all tests

puneetdixit200:
## Summary
- stop making the default event_loop_policy fixture autouse
- request event_loop_policy only for pytest-asyncio tests and tests using asyncio fixtures, so plain sync tests are not parametrized
- add regression coverage for plain sync tests and sync tests that use asyncio fixtures
- avoid re-importing pytest_asyncio in one subprocess test under -W error; the entry point already loads the plugin

Fixes #796

## Tests
- .venv/bin/python -m pytest -q
- .venv/bin/python -m pytest tests/test_set_event_loop.py::test_asyncio_run_after_async_fixture_does_not_leak_loop tests/markers/test_function_scope.py::test_parametrized_loop_policy_does_not_parametrize_sync_tests tests/markers/test_function_scope.py::test_parametrized_loop_policy_parametrizes_sync_tests_with_async_fixtures -q
- .venv/bin/python -m pytest tests/markers/test_function_scope.py tests/markers/test_module_scope.py tests/markers/test_class_scope.py tests/markers/test_session_scope.py tests/test_loop_factory_parametrization.py tests/test_asyncio_fixture.py -q
- .venv/bin/python -m ruff check pytest_asyncio/plugin.py tests/markers/test_function_scope.py tests/test_set_event_loop.py
- .venv/bin/python -m mypy --python-version 3.13 pytest_asyncio/
- .venv/bin/python -m pyright --pythonpath .venv/bin/python --pythonversion 3.13 pytest_asyncio/
- git diff --check

### DIFF
--- a/pytest_asyncio/plugin.py
+++ b/pytest_asyncio/plugin.py
@@ -795,14 +795,16 @@ def pytest_generate_tests(metafunc: pytest.Metafunc) -> None:
 
 
 def _add_fixture_to_metafunc(metafunc: pytest.Metafunc, fixture_name: str) -> None:
-    if fixture_name not in metafunc.fixturenames:
-        metafunc.fixturenames.append(fixture_name)
-
-    if fixture_name not in metafunc._arg2fixturedefs:
-        fixturemanager = metafunc.definition.session._fixturemanager
-        fixturedefs = fixturemanager.getfixturedefs(fixture_name, metafunc.definition)
-        if fixturedefs is not None:
-            metafunc._arg2fixturedefs[fixture_name] = fixturedefs
+    fixturemanager = metafunc.definition.session._fixturemanager
+    fixturenames_closure, arg2fixturedefs = fixturemanager.getfixtureclosure(
+        metafunc.definition,
+        (fixture_name,),
+        ignore_args=set(),
+    )
+    metafunc._arg2fixturedefs.update(arg2fixturedefs)
+    for name in fixturenames_closure:
+        if name not in metafunc.fixturenames:
+            metafunc.fixturenames.append(name)
 
 
 def _uses_asyncio_fixtures(metafunc: pytest.Metafunc) -> bool:
--- a/pytest_asyncio/plugin.py
+++ b/pytest_asyncio/plugin.py
@@ -729,14 +729,23 @@ def pytest_generate_tests(metafunc: pytest.Metafunc) -> None:
         metafunc.definition
     )
     if specialized_item_class is None:
+        if _uses_asyncio_fixtures(metafunc):
+            _add_fixture_to_metafunc(metafunc, "event_loop_policy")
         return
 
     asyncio_marker = _resolve_asyncio_marker(metafunc.definition)
     if asyncio_marker is None:
+        if _uses_asyncio_fixtures(metafunc):
+            _add_fixture_to_metafunc(metafunc, "event_loop_policy")
         return
     marker_loop_scope, marker_selected_factory_names = _parse_asyncio_marker(
         asyncio_marker
     )
+    default_loop_scope = _get_default_test_loop_scope(metafunc.config)
+    loop_scope = marker_loop_scope or default_loop_scope
+    runner_fixture_id = f"_{loop_scope}_scoped_runner"
+    _add_fixture_to_metafunc(metafunc, runner_fixture_id)
+    _add_fixture_to_metafunc(metafunc, "event_loop_policy")
 
     hook_factories = _collect_hook_loop_factories(metafunc.config, metafunc.definition)
     if hook_factories is None:
@@ -774,8 +783,6 @@ def pytest_generate_tests(metafunc: pytest.Metafunc) -> None:
             for name in marker_selected_factory_names
         ]
     metafunc.fixturenames.append(_asyncio_loop_factory.__name__)
-    default_loop_scope = _get_default_test_loop_scope(metafunc.config)
-    loop_scope = marker_loop_scope or default_loop_scope
     # pytest.HIDDEN_PARAM was added in pytest 8.4
     hide_id = len(factory_ids) == 1 and hasattr(pytest, "HIDDEN_PARAM")
     metafunc.parametrize(
@@ -785,6 +792,30 @@ def pytest_generate_tests(metafunc: pytest.Metafunc) -> None:
         indirect=True,
         scope=loop_scope,
     )
+
+
+def _add_fixture_to_metafunc(metafunc: pytest.Metafunc, fixture_name: str) -> None:
+    if fixture_name not in metafunc.fixturenames:
+        metafunc.fixturenames.append(fixture_name)
+
+    if fixture_name not in metafunc._arg2fixturedefs:
+        fixturemanager = metafunc.definition.session._fixturemanager
+        fixturedefs = fixturemanager.getfixturedefs(fixture_name, metafunc.definition)
+        if fixturedefs is not None:
+            metafunc._arg2fixturedefs[fixture_name] = fixturedefs
+
+
+def _uses_asyncio_fixtures(metafunc: pytest.Metafunc) -> bool:
+    asyncio_mode = _get_asyncio_mode(metafunc.config)
+    for fixturedefs in metafunc._arg2fixturedefs.values():
+        fixturedef = fixturedefs[-1]
+        fixture_func = fixturedef.func
+        if _is_asyncio_fixture_function(fixture_func):
+            return True
+        if asyncio_mode == Mode.AUTO and _is_coroutine_or_asyncgen(fixture_func):
+            return True
+
+    return False
 
 
 @contextlib.contextmanager
@@ -1073,7 +1104,7 @@ def _asyncio_loop_factory(request: FixtureRequest) -> LoopFactory | None:
     return getattr(request, "param", None)
 
 
-@pytest.fixture(scope="session", autouse=True)
+@pytest.fixture(scope="session")
 def event_loop_policy() -> AbstractEventLoopPolicy:
     """Return an instance of the policy used to create asyncio event loops."""
     return _get_event_loop_policy()
--- a/pytest_asyncio/plugin.py
+++ b/pytest_asyncio/plugin.py
@@ -724,6 +724,36 @@ def pytest_pycollect_makeitem_convert_async_functions_to_subclass(
 
 
 @pytest.hookimpl(tryfirst=True)
+def pytest_collection_modifyitems(items: list[pytest.Item]) -> None:
+    for item in items:
+        if _item_needs_event_loop_policy(item):
+            _append_fixture_name(item, "event_loop_policy")
+
+
+def _item_needs_event_loop_policy(item: pytest.Item) -> bool:
+    if not isinstance(item, Function):
+        return False
+    if isinstance(item, PytestAsyncioFunction) and _resolve_asyncio_marker(item):
+        return True
+    return _item_uses_asyncio_fixtures(item)
+
+
+def _item_uses_asyncio_fixtures(item: pytest.Item) -> bool:
+    fixtureinfo = getattr(item, "_fixtureinfo", None)
+    if fixtureinfo is None:
+        return False
+    return _fixture_defs_use_asyncio(
+        item.config,
+        fixtureinfo.name2fixturedefs.values(),
+    )
+
+
+def _append_fixture_name(item: pytest.Item, fixture_name: str) -> None:
+    if fixture_name not in item.fixturenames:
+        item.fixturenames.append(fixture_name)
+
+
+@pytest.hookimpl(tryfirst=True)
 def pytest_generate_tests(metafunc: pytest.Metafunc) -> None:
     specialized_item_class = PytestAsyncioFunction.item_subclass_for(
         metafunc.definition
@@ -808,8 +838,20 @@ def _add_fixture_to_metafunc(metafunc: pytest.Metafunc, fixture_name: str) -> No
 
 
 def _uses_asyncio_fixtures(metafunc: pytest.Metafunc) -> bool:
-    asyncio_mode = _get_asyncio_mode(metafunc.config)
-    for fixturedefs in metafunc._arg2fixturedefs.values():
+    return _fixture_defs_use_asyncio(
+        metafunc.config,
+        metafunc._arg2fixturedefs.values(),
+    )
+
+
+def _fixture_defs_use_asyncio(
+    config: Config,
+    fixturedefs_by_name: Iterable[Sequence[FixtureDef[Any]]],
+) -> bool:
+    asyncio_mode = _get_asyncio_mode(config)
+    for fixturedefs in fixturedefs_by_name:
+        if not fixturedefs:
+            continue
         fixturedef = fixturedefs[-1]
         fixture_func = fixturedef.func
         if _is_asyncio_fixture_function(fixture_func):
--- a/tests/markers/test_function_scope.py
+++ b/tests/markers/test_function_scope.py
@@ -217,6 +217,35 @@ def test_parametrized_loop_policy_parametrizes_sync_tests_with_async_fixtures(
     result.assert_outcomes(passed=2)
 
 
+def test_parametrized_loop_policy_can_depend_on_parametrized_fixture(
+    pytester: Pytester,
+):
+    pytester.makeini("[pytest]\nasyncio_default_fixture_loop_scope = function")
+    pytester.makepyfile(dedent("""\
+            import asyncio
+
+            import pytest
+
+            @pytest.fixture(params=["policy_a", "policy_b"])
+            def policy(request):
+                return request.param
+
+            @pytest.fixture
+            def event_loop_policy(policy):
+                assert policy in {"policy_a", "policy_b"}
+                return asyncio.get_event_loop_policy()
+
+            @pytest.mark.asyncio
+            async def test_marked_async():
+                pass
+
+            async def test_auto_async():
+                pass
+            """))
+    result = pytester.runpytest("--asyncio-mode=auto")
+    result.assert_outcomes(passed=4)
+
+
 def test_event_loop_policy_fixture_override_emits_deprecation_warning(
     pytester: Pytester,
 ):
--- a/tests/markers/test_function_scope.py
+++ b/tests/markers/test_function_scope.py
@@ -217,6 +217,74 @@ def test_parametrized_loop_policy_parametrizes_sync_tests_with_async_fixtures(
     result.assert_outcomes(passed=2)
 
 
+def test_parametrized_loop_policy_parametrizes_auto_mode_async_fixtures(
+    pytester: Pytester,
+):
+    pytester.makeini("[pytest]\nasyncio_default_fixture_loop_scope = function")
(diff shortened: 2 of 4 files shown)


golikovichev (commented):
Approach looks clean. Dropping `autouse=True` from the `event_loop_policy` fixture and gating its injection on `pytest_generate_tests` only when the test uses asyncio fixtures (or is async itself) is the right shape to fix #796 without breaking the loop-policy parametrization use case.

Two observations from a downstream-user read:

1. `_uses_asyncio_fixtures` walks `metafunc._arg2fixturedefs` and short-circuits on first match. In AUTO mode it falls back to `_is_coroutine_or_asyncgen`, which is what catches sync tests that happen to request async fixtures - exactly the gap #796 surfaced. Worth keeping the AUTO branch as-is even though it adds a per-fixture check.
2. The negative-case regression test `test_parametrized_loop_policy_does_not_parametrize_sync_tests` is paired with `test_parametrized_loop_policy_parametrizes_sync_tests_with_async_fixtures` (positive case). Good coverage pattern.

CI green across the full matrix (Py 3.10-3.14 + Windows + pytest main). Lint + mypy + pyright also clean per PR body.

Not a maintainer, leaving this as a non-blocking review.

pytest_asyncio/plugin.py
```diff
@@ -787,6 +794,30 @@ def pytest_generate_tests(metafunc: pytest.Metafunc) -> None:
     )
 
 
+def _add_fixture_to_metafunc(metafunc: pytest.Metafunc, fixture_name: str) -> None:
```
tjkuson, line 827: I think this changes pytest's internals and breaks valid usage. For example, on main, this passes.

```
$ cat >mre.py <<'# EOF'
import asyncio
import pytest


@pytest.fixture(params=[asyncio.DefaultEventLoopPolicy(), asyncio.DefaultEventLoopPolicy()])
def policy(request):
    return request.param


@pytest.fixture
def event_loop_policy(policy):
    return policy


@pytest.mark.asyncio
async def test_async():
    pass


async def test_sync():
    pass
# EOF

$ pytest mre.py
==================================================== test session starts ====================================================
platform darwin -- Python 3.13.0, pytest-9.0.3, pluggy-1.6.0
rootdir: /Users/tjkuson/workspace/pytest-asyncio
configfile: pyproject.toml
plugins: asyncio-1.4.0a3.dev24
asyncio: mode=Mode.AUTO, debug=False, asyncio_default_fixture_loop_scope=function, asyncio_default_test_loop_scope=function
collected 2 items

mre.py EE                                                                                                             [100%]

========================================================== ERRORS ===========================================================
_______________________________________________ ERROR at setup of test_async ________________________________________________
The requested fixture has no parameter defined for test:
    mre.py::test_async

Requested fixture 'policy' defined in:
mre.py:6

Requested here:
.venv/lib/python3.13/site-packages/_pytest/fixtures.py:627
________________________________________________ ERROR at setup of test_sync ________________________________________________
The requested fixture has no parameter defined for test:
    mre.py::test_sync

Requested fixture 'policy' defined in:
mre.py:6

Requested here:
.venv/lib/python3.13/site-packages/_pytest/fixtures.py:627
===================================================== 2 errors in 0.02s =====================================================
```

golikovichev (commented):
Good catch with the MRE. My read was too superficial - I focused on the `autouse` change and the regression test pair, but missed that `_add_fixture_to_metafunc` injecting `event_loop_policy` breaks valid fixture chains where `event_loop_policy` depends on another parametrized fixture (like your `policy`). The patch shifts the parametrization origin out from under pytest's resolver.

Leaving this to the PR author and maintainers.

golikovichev (commented):
Re-reviewed after `8305b3a`. The new `_add_fixture_to_metafunc` walks the closure via `fixturemanager.getfixtureclosure()` instead of attaching just the named fixture, so parametrize chains like `event_loop_policy(policy)` from @tjkuson's MRE are now preserved end to end.

The new test `test_parametrized_loop_policy_can_depend_on_parametrized_fixture` mirrors the MRE shape directly (parametrized `policy` fixture, `event_loop_policy(policy)` depending on it, both async and sync tests), so the regression is pinned. The two surrounding sync-vs-async parametrization tests cover the adjacent cases I would have asked for, so I do not have anything to add.

Sorry again for missing this in my first read - your MRE made the failure mode obvious, and the fix landed cleanly.

gmail-precheck-done

puneetdixit200:
Pushed cd89044 to cover the remaining event loop policy fixture discovery paths from the Codecov report.

Added coverage for auto-mode regular async fixtures and strict-mode unmarked async tests that still need the policy fixture closure.

Verified locally:
- `coverage run -m pytest -q` (`304 passed`)
- `coverage report -m pytest_asyncio/plugin.py` (the previously missing new lines are now covered)
- `python -m pytest tests/markers/test_function_scope.py tests/test_set_event_loop.py -q` (`78 passed`)
- `ruff check tests/markers/test_function_scope.py`
- `git diff --check`

tests/markers/test_function_scope.py
```diff
@@ -154,6 +154,166 @@ async def test_parametrized_loop():
...
+
+            import pytest
+
+            @pytest.fixture(params=["policy_a", "policy_b"])
+            def policy(request):
+                return request.param
+
+            @pytest.fixture
+            def event_loop_policy(policy):
+                assert policy in {"policy_a", "policy_b"}
```
seifertm: I don't see any benefit of this statement.

```suggestion
```

tests/markers/test_function_scope.py
```diff
@@ -154,6 +154,166 @@ async def test_parametrized_loop():
...
+
+            @pytest_asyncio.fixture
+            async def async_fixture():
+                return True
+
+            async def test_unmarked_async(async_fixture):
+                pass
+            """))
+    result = pytester.runpytest("--asyncio-mode=strict")
+    result.assert_outcomes(failed=2)
```
seifertm: This test case parametrizes the loop policy and provides an async fixture. Pytest-asyncio is run in *strict* mode and *test_unmarked_async* is not marked, so it should never execute in my understanding.
 
Can you explain (1) why we expect two test cases to run, (2) why they should fail, and how the test contributes to the functionality of this PR?

tests/markers/test_function_scope.py
seifertm, line 1: The added tests have similar structure, but often contain superfluous code. For example, some tests parametrize the loop policy with `asyncio.get_event_loop_policy()`, others create custom policy implementations. Subsequent tests assert that the current policy's name is the one expected from the parametrization, etc. For this PR, there's no need for tests inside Pytester to make any kind assertions. I find them distracting when trying to understanding the actual purpose of the tests.

I'll leave suggestions in the relevant places and I appreciate if you changed the tests in the spirit of this comment.

tests/markers/test_function_scope.py
```diff
@@ -154,6 +154,166 @@ async def test_parametrized_loop():
...
+            )
+            def event_loop_policy(request):
+                return request.param
+
+            @pytest_asyncio.fixture
+            async def async_fixture():
+                return True
+
+            def test_sync_with_async_fixture(async_fixture):
+                assert async_fixture
```
seifertm:
```suggestion
                pass
```

tests/markers/test_function_scope.py
```diff
@@ -154,6 +154,166 @@ async def test_parametrized_loop():
...
+                    asyncio.get_event_loop_policy(),
+                ],
+                ids=["policy_a", "policy_b"],
+            )
+            def event_loop_policy(request):
+                return request.param
+
+            @pytest_asyncio.fixture
+            async def async_fixture():
+                return True
```
seifertm:
```suggestion
                pass
```

tests/markers/test_function_scope.py
```diff
@@ -154,6 +154,166 @@ async def test_parametrized_loop():
...
+    pytester.makeini("[pytest]\nasyncio_default_fixture_loop_scope = function")
+    pytester.makepyfile(dedent("""\
+            import asyncio
+
+            import pytest
+
+            class CustomEventLoopPolicy(asyncio.DefaultEventLoopPolicy):
+                def __init__(self, name):
+                    super().__init__()
+                    self.name = name
```
seifertm:
```suggestion
```

tests/markers/test_function_scope.py
```diff
@@ -154,6 +154,166 @@ async def test_parametrized_loop():
...
+
+            @pytest.fixture(
+                scope="session",
+                params=[
+                    CustomEventLoopPolicy("policy_a"),
+                    CustomEventLoopPolicy("policy_b"),
```
seifertm:
```suggestion
                    asyncio.get_event_loop_policy(),
                    asyncio.get_event_loop_policy(),
```

tests/markers/test_function_scope.py
```diff
@@ -154,6 +154,166 @@ async def test_parametrized_loop():
...
+                ],
+                ids=["policy_a", "policy_b"],
+            )
+            def event_loop_policy(request):
+                return request.param
+
+            @pytest.fixture
+            async def async_fixture():
+                return asyncio.get_event_loop_policy().name
```
seifertm:
```suggestion
                pass
```

tests/markers/test_function_scope.py
```diff
@@ -154,6 +154,166 @@ async def test_parametrized_loop():
...
+
+            def test_sync_with_auto_async_fixture(async_fixture):
+                assert async_fixture in {"policy_a", "policy_b"}
```
seifertm:
```suggestion
                pass
```

tests/markers/test_function_scope.py
```diff
@@ -154,6 +154,166 @@ async def test_parametrized_loop():
...
+                    asyncio.get_event_loop_policy(),
+                ],
+                ids=["policy_a", "policy_b"],
+            )
+            def event_loop_policy(request):
+                return request.param
+
+            @pytest_asyncio.fixture
+            async def async_fixture():
+                return True
```
seifertm:
```suggestion
                pass
```

pytest_asyncio/plugin.py
seifertm: @tjkuson Have your concerns been addressed by the latest commit? If not, can you elaborate?

puneetdixit200: This is covered on the current head `cd89044`: `_add_fixture_to_metafunc()` now asks pytest for the full fixture closure with `getfixtureclosure()` before appending fixture names, so `event_loop_policy(policy)` carries the parametrized `policy` dependency through.

The regression test `test_parametrized_loop_policy_can_depend_on_parametrized_fixture` mirrors this MRE shape directly. I rechecked locally:
- `.venv/bin/python -m pytest tests/markers/test_function_scope.py::test_parametrized_loop_policy_can_depend_on_parametrized_fixture -q` -> `1 passed`
- `.venv/bin/python -m pytest tests/markers/test_function_scope.py tests/test_set_event_loop.py -q` -> `78 passed`
- `.venv/bin/python -m ruff check pytest_asyncio/plugin.py tests/markers/test_function_scope.py tests/test_set_event_loop.py` -> `All checks passed!`

seifertm (changes requested):
@puneetdixit200 I appreciate the initiative. The proposed changes look reasonable. The PR solves a long-standing issue in pytest-asyncio.

I wish you had reached out before authoring the PR, though. The policy system has been deprecated in Python and will consequently be deprecated in pytest-asyncio. The latest pytest-asyncio version (v1.4.0a2, stable release today) already contains an approach that allows parametrizing tests with different loops, leaving the policy system deprecated. I'm not saying we cannot merge this, but I think your efforts would have been better spent on a different issue.

If you could have a look at my comments, that would be great  :) I suggest to first read through my text comments and leave the code suggestions for last

seifertm:
@golikovichev I *do* appreciate reviews from non-maintainers. Thank you.

tests/markers/test_function_scope.py
puneetdixit200: Adjusted in `4ec829c`:

- removed the strict-mode unmarked async test that did not exercise pytest-asyncio behavior clearly
- simplified the Pytester bodies so they only assert collection/outcome counts instead of asserting fixture return values inside the generated tests
- replaced the custom named loop policy class in the auto-mode fixture test with plain policy instances

Local validation:
- `.venv/bin/python -m pytest tests/markers/test_function_scope.py::test_parametrized_loop_policy_parametrizes_sync_tests_with_async_fixtures tests/markers/test_function_scope.py::test_parametrized_loop_policy_parametrizes_auto_mode_async_fixtures tests/markers/test_function_scope.py::test_parametrized_loop_policy_can_depend_on_parametrized_fixture -q` -> `3 passed`
- `.venv/bin/python -m pytest tests/markers/test_function_scope.py tests/test_set_event_loop.py -q` -> `77 passed`
- `.venv/bin/python -m ruff check pytest_asyncio/plugin.py tests/markers/test_function_scope.py tests/test_set_event_loop.py` -> `All checks passed!`
- `git diff --check`

puneetdixit200:
Pushed `63cbfa3` for the review feedback.

Changes:
- Removed the strict-mode unmarked async branch from `pytest_generate_tests`; unmarked async tests are not managed by pytest-asyncio in strict mode, so they should not get `event_loop_policy` fixture injection.
- Removed the remaining custom IDs from the Pytester loop policy fixtures.
- Kept the simplified Pytester bodies from the prior cleanup commit.

Verification:
- `UV_CACHE_DIR=/tmp/pytest-asyncio-1454-uv-cache uv run --extra testing python -m pytest tests/markers/test_function_scope.py tests/test_set_event_loop.py -q` (`77 passed`)
- `UV_CACHE_DIR=/tmp/pytest-asyncio-1454-uv-cache uv run --with ruff ruff check pytest_asyncio/plugin.py tests/markers/test_function_scope.py`
- `UV_CACHE_DIR=/tmp/pytest-asyncio-1454-uv-cache uv run --extra testing coverage run -m pytest -q` (`303 passed`)
- `UV_CACHE_DIR=/tmp/pytest-asyncio-1454-uv-cache uv run --extra testing coverage report -m pytest_asyncio/plugin.py`
- `git diff --check`

puneetdixit200:
Follow-up on the latest push: current head `e450cc8` supersedes my earlier `63cbfa3` note. This head is test-only and leaves the plugin implementation from `cd89044` intact while simplifying the review-covered Pytester cases.

Verified locally before pushing:
- `tests/markers/test_function_scope.py -q` (`16 passed`)
- `tests/test_set_event_loop.py -q` (`61 passed`)
- `ruff check tests/markers/test_function_scope.py`
- `git diff --check`

The GitHub check suite is now green on `e450cc8` as well.

pytest_asyncio/plugin.py
tjkuson: @seifertm The specific issue is addressed, but I still feel a bit concerned about the intervention in pytest's internals and possible breakage for users. For instance, `event_loop_policy` injected late, meaning users can no longer do this:

```python
def pytest_collection_modifyitems(items):
    for item in items:
        if "event_loop_policy" in item.fixturenames:
            item.add_marker(pytest.mark.uses_loop_policy)
```

It's admittedly low-level and I doubt many users are doing things like this, but it's still IMO valid pytest usage of a documented and public API. I don't have any ideas at the moment about how to resolve this.

puneetdixit200: Thanks, this is a valid compatibility concern. I pushed `2bdec76` to keep the injected `event_loop_policy` name visible before user `pytest_collection_modifyitems` hooks run, while still avoiding the old autouse behavior for plain sync tests.

I added `test_injected_loop_policy_is_visible_during_collection_modifyitems`, which mirrors the hook pattern from your example: marked async items get the marker via `item.fixturenames`, and the plain sync test does not.

Local verification before pushing:
- `UV_CACHE_DIR=/private/tmp/pytest_asyncio_1454_uv_cache uv run --extra testing python -m pytest tests/markers/test_function_scope.py::test_injected_loop_policy_is_visible_during_collection_modifyitems -q` -> `1 passed`
- `UV_CACHE_DIR=/private/tmp/pytest_asyncio_1454_uv_cache uv run --extra testing python -m pytest tests/markers/test_function_scope.py -q` -> `17 passed`
- `UV_CACHE_DIR=/private/tmp/pytest_asyncio_1454_uv_cache uv run --extra testing python -m pytest tests/test_set_event_loop.py -q` -> `61 passed`
- `UV_CACHE_DIR=/private/tmp/pytest_asyncio_1454_uv_cache uv run --extra testing python -m pytest -q` -> `304 passed`
- `UV_CACHE_DIR=/private/tmp/pytest_asyncio_1454_uv_cache uvx ruff check pytest_asyncio/plugin.py tests/markers/test_function_scope.py` -> `All checks passed!`
- `git diff --check`

golikovichev:
Thanks @seifertm, good to hear. Nice to see #1346 wrapped up via #1373 in the same window too.

seifertm:
The author seems unwilling to answer [questions](https://github.com/pytest-dev/pytest-asyncio/pull/1454#discussion_r3302392261) about the submitted code. They resolve [review comments](https://github.com/pytest-dev/pytest-asyncio/pull/1454#discussion_r3284739850) prematurely, instead of engaging in a discussion about the underlying problem. At the same time, they don't interact with the many code suggestions provided by reviewers at all and seem to [push code](https://github.com/pytest-dev/pytest-asyncio/pull/1454#discussion_r3303448246) for no obvious reason.

My impression is that I'm not talking to a human, but solely to a bot. I go out of my way to pretend this is not the case, since OP's account is not labelled as such. There appears to be no consideration whatsoever for the time invested by maintainers. The wanton disregard for other people's time–by this PR and other recent ones–reached a point where I find this behavior and communication style personally offensive.

I'm simply unwilling to invest further time into this PR.

Pytest-asyncio needs a contribution guide.

PR #1462, opened 2026-05-30
fix: default asyncio_default_fixture_loop_scope to "function"

### LINKED ISSUE #924: Finish deprecation of unset asyncio_default_fixture_loop_scope

sridhar-3009:
## Summary

Closes #924.

The `asyncio_default_fixture_loop_scope` ini option was left `None` by default, which triggered a `PytestDeprecationWarning` on every test run and caused async fixtures to fall back to the old fixture-caching scope behaviour.

This PR completes the deprecation by:

1. **Setting the default to `"function"`** — the value the deprecation warning was telling users to set explicitly.
2. **Removing `_DEFAULT_FIXTURE_LOOP_SCOPE_UNSET`** — the warning constant is now unreachable since the default is no longer falsy.
3. **Removing the `if not default_fixture_loop_scope: warnings.warn(...)` block** in `pytest_configure`.

## Before

```ini
# No ini setting needed, but you got a warning every run:
# PytestDeprecationWarning: The configuration option
#   "asyncio_default_fixture_loop_scope" is unset. ...
```

## After

```ini
# No ini setting needed, no warning.
# Fixtures default to "function" scope.
```

## Follow-up

Tests that explicitly set `asyncio_default_fixture_loop_scope = function` in pytester configs are now setting the default redundantly. Cleaning those up is a separate, mechanical follow-up.

### DIFF
--- a/changelog.d/924.changed.rst
+++ b/changelog.d/924.changed.rst
@@ -1,0 +1,3 @@
+``asyncio_default_fixture_loop_scope`` now defaults to ``"function"`` instead of being unset.
+Previously, leaving the option unset caused a ``PytestDeprecationWarning`` on every test run.
+The old deprecation warning and its associated constant have been removed.
--- a/pytest_asyncio/plugin.py
+++ b/pytest_asyncio/plugin.py
@@ -271,8 +271,6 @@ def _collect_hook_loop_factories(
     return factories
 
 
-
-
 def _validate_scope(scope: str | None, option_name: str) -> None:
     if scope is None:
         return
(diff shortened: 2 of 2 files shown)


golikovichev:
Quick heads-up: there are two other open PRs against the same issue (#924):

- #1444 by @SahilArchitect - set default to `"function"` and adds a regression test that asserts the warning is gone. Already has maintainer review feedback applied. Most recent activity was 2026-05-15, looks close to landing.
- #1381 by @JiwaniZakir - older single-line default change, no test, no human review yet.

This PR is closest in shape to #1381 (drop the warning block, set the default), and the diff overlaps almost entirely with the plugin.py change in #1444. To keep things tidy for the maintainers, would you be open to either:

1. Closing this PR in favour of #1444 and dropping a review there (an "approves" or a concrete comment on the regression test would probably help that one land faster), or
2. Rebasing here on top of #1444 once it merges, if there's something left to do.

If you spotted a behaviour gap that #1444 doesn't cover, that's worth flagging on the issue thread so all three PRs converge on the same target rather than triaging the same default change three times.

sridhar-3009:
Closing in favour of #1444 which is further along and has maintainer feedback applied. Will follow that PR instead.

PR #1473, opened 2026-06-04
Deprecate @pytest_asyncio.fixture for synchronous fixtures

### LINKED ISSUE #1090: Type error in `pytest-asyncio==0.26.0` with: `Value of type variable "_R" of function cannot be "AsyncClient"`
Upgrading to `pytest-asyncio==0.26.0`:

```python
import pytest_asyncio
from httpx import AsyncClient


@pytest_asyncio.fixture(scope="function")
def server_async_client() -> AsyncClient: ...
```

I get this error from `mypy`:

```none
a.py:5:2: error: Value of type variable "_R" of function cannot be "AsyncClient"  [type-var]
    @pytest_asyncio.fixture(scope="function")
     ^
```

This is with Python 3.12.7 and the following packages:

```none
httpx                     0.27.2
mypy                      1.15.0
pytest                    8.3.5
pytest-asyncio            0.25.3
```

SaInekK:
Using `@pytest_asyncio.fixture` on a synchronous function is a user mistake — synchronous fixtures should use `@pytest.fixture`. Previously this silently succeeded, making the error hard to diagnose.

This PR:
- Widens `_R = TypeVar("_R")` (removes the `Awaitable | AsyncIterator` bound) so applying `@pytest_asyncio.fixture` to a sync function no longer raises a type error before the warning fires.
- Emits `PytestDeprecationWarning` at decoration time when the decorated function is not a coroutine or async generator.
- Adds tests for both the bare-decorator and factory-decorator paths.

Closes #1090

### DIFF
--- a/changelog.d/1090.deprecated.rst
+++ b/changelog.d/1090.deprecated.rst
@@ -1,0 +1,1 @@
+Using :func:`pytest_asyncio.fixture` to decorate synchronous fixtures is now deprecated. Use :func:`pytest.fixture` instead.
--- a/pytest_asyncio/plugin.py
+++ b/pytest_asyncio/plugin.py
@@ -14,8 +14,6 @@ import traceback
 import warnings
 from asyncio import AbstractEventLoop
 from collections.abc import (
-    AsyncIterator,
-    Awaitable,
     Callable,
     Collection,
     Generator,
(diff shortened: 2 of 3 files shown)


seifertm:
The PR relaxes type constraints on `pytest_asyncio.fixture` in favor of raising a *PytestDeprecationWarning*. Consequently, synchronous fixtures decorated with `pytest_asyncio.fixture` no longer result in a warning from the type checker, but instead emit a warning at runtime.

While the submitted patch technically solves the linked issue, the chosen approach defeats the purpose of typed code, which tries to catch issues in a static analysis step as opposed to at runtime. Users will also lose typing errors for the deprecated use in their IDEs and editors. Unless you can convince me otherwise, I'll close this PR.

PR #1486, opened 2026-06-20
Remove deprecated `scope` keyword argument from asyncio marker

### LINKED ISSUE #1475: Turn deprecation of the "scope" keyword argument of the asyncio marker into an error
Pytest-asyncio deprecated the `scope` keyword argument of the `asyncio` marker in favor of `loop_scope` back in 2024 (commit a40333a). Passing `scope=...` to `@pytest.mark.asyncio` currently emits a `PytestDeprecationWarning` and is silently treated as `loop_scope`.

The goal of this issue is to follow through with the deprecation and convert the warning into an error, consistent with #1202 and #924.

This is a breaking change that involves a major version bump to pytest-asyncio.

For reference, the relevant code lives in `_parse_asyncio_marker` (`pytest_asyncio/plugin.py`), and the behavior is currently covered in `tests/markers/test_function_scope.py`. The duplicate `scope` + `loop_scope` case already raises `UsageError`, so the error path is partially in place.

Would this fit into the v2.0 milestone alongside the other deprecation follow-throughs? If so, I'd be happy to pick it up.

mokashang:
Fixes #1475.

## What changed

The `scope` keyword argument to `@pytest.mark.asyncio` was deprecated in v0.23 in favor of `loop_scope`. Per the v2.0 milestone discussion on #1475, this PR converts the deprecation warning into a `pytest.UsageError`.

It takes the broader approach @seifertm suggested in the issue: any unrecognized keyword argument to the marker now raises a usage error, with no special-case handling for `scope` left in the parser. The error message names the offending kwargs by repr (e.g. `'scope'`, `'cope'`) rather than printing a generic "only `loop_scope` and `loop_factories` are accepted" sentence. Positional-argument validation is split into its own branch so its message can mention positionals specifically rather than lumping them in with unknown kwargs.

## Surface-level details

- `_validate_asyncio_marker` now raises `pytest.UsageError` (was `ValueError`); the kwarg allow-set is `{"loop_scope", "loop_factories"}` with no `"scope"`.
- `_parse_asyncio_marker` loses the `if "scope" in kwargs:` branch and the duplicate-loop_scope error path (which is now unreachable, since `scope` can no longer reach this function at all).
- `PytestAsyncioFunction._loop_scope` drops the `or marker.kwargs.get("scope")` fallback.
- The `_MARKER_SCOPE_KWARG_DEPRECATION_WARNING` and `_DUPLICATE_LOOP_SCOPE_DEFINITION_ERROR` strings are deleted.
- The v0.23 migration guide is updated to note that `scope=` is now an error rather than a warning.

## Tests

- `tests/markers/test_function_scope.py::test_warns_when_scope_argument_is_present` removed; the existing `test_raises_when_scope_and_loop_scope_arguments_are_present` is extended with an `fnmatch_lines` assertion on the new "unexpected keyword argument 'scope'" message so the regression is pinned to the new error path.
- `tests/markers/test_invalid_arguments.py`: the three `ValueError`-asserting tests are updated to match the new `UsageError` message wording, and a new `test_error_when_scope_keyword_argument_is_passed` covers the lone-`scope=` case that was formerly only a deprecation warning.
- The targeted suite (`tests/markers/test_function_scope.py` + `tests/markers/test_invalid_arguments.py`, 20 tests) passes locally. The remaining 27 failures in `python -m pytest` on `main` are pre-existing subprocess-launch failures in my local environment and reproduce verbatim against an unmodified `main` checkout, so they're unrelated to this change.

## Changelog

Added `changelog.d/1475.removed.rst` under the `Removed` category, since the kwarg is being removed in v2.0.

## Notes

I deliberately left `_INVALID_LOOP_FACTORIES_KWARG` raising `ValueError` rather than `UsageError` — that's a value-shape check, not a kwarg-name check, and converting it felt out of scope for this issue. Happy to fold it in if you'd prefer the validators speak the same exception type.


SaInekK:
Thanks for taking this on, @mokashang — the approach matches what @seifertm and I landed on in #1475 (UsageError for any unknown kwarg, dropping the `scope` special-casing), so this reads right to me.

One CI nit: the **Run linters** job is red on the `shed` hook, which wants to reformat `tests/markers/test_invalid_arguments.py`. Running `pre-commit run --all-files` locally should auto-fix it and turn the job green.

Splitting the positional-arg error out from the unknown-kwarg path is a nice touch. 👍

mokashang:
Thanks @SaInekK — good catch. Pushed 53cfd58 collapsing the `fnmatch_lines` call onto one line; `pre-commit run --all-files` is clean locally and all 9 tests in `test_invalid_arguments.py` (plus the 11 in `test_function_scope.py`) still pass.

PR #1502, opened 2026-07-13
Fix loop factory teardown ordering

### LINKED ISSUE #1501: Loop factory parametrization tears down async fixtures used by sync tests
I encountered an issue with defining `pytest_asyncio_loop_factories` in test suites where non-async tests use async fixtures in shared loop scopes.

To reproduce:

```python
# conftest.py

import asyncio
from collections.abc import Mapping, Callable

import pytest


def pytest_asyncio_loop_factories(
    config: pytest.Config,
    item: pytest.Item,
) -> Mapping[str, Callable[[], asyncio.AbstractEventLoop]]:
    return {"default": asyncio.new_event_loop}
```

```python
# test_mre.py

import pytest
import pytest_asyncio


@pytest_asyncio.fixture(scope="session")
async def parent() -> AsyncGenerator[str]:
    yield "parent"


@pytest_asyncio.fixture(scope="session")
async def child(parent: str) -> AsyncGenerator[str]:
    yield "child"


@pytest.mark.asyncio(loop_scope="session")
async def test_async(parent) -> None:
    assert parent == "parent"


def test_sync(child: str) -> None:
    assert child == "child"
```

Running `test_mre.py` on the above using pytest-asyncio 1.4.0 and pytest 9.1.1 shows one pass and one failure with the following errror:

```
AssertionError: The fixture value for "parent" is not available.  This can happen when the fixture has already been torn down.
```

If `pytest_asyncio_loop_factories` is not defined in the conftest file, both tests pass. 

If `test_async` is after `test_sync`, both tests pass.

The expected behaviour is that both tests pass regardless of order and `pytest_asyncio_loop_factories` hook implementation.

I think the fix would involve handling the loop factory for any test item that uses a pytest-asyncio-managed async fixture, not just async tests themselves.

Sanjays2402:
Fixes #1501. Loop-factory parametrization changed the scoped runner without recording it as a fixture dependency, so pytest finalized a shared parent fixture before a later synchronous test initialized its child. Recording that dependency preserves pytest's normal child-before-parent teardown order; the new regression test fails with the reported assertion before the fix and passes after it.

### DIFF
--- a/changelog.d/1501.fixed.rst
+++ b/changelog.d/1501.fixed.rst
@@ -1,0 +1,1 @@
+The scoped runner is now registered as a dependency of managed fixtures. This ensures loop factory parameter changes tear down dependent fixtures in dependency order. (`#1501 <https://github.com/pytest-dev/pytest-asyncio/issues/1501>`_)
--- a/pytest_asyncio/plugin.py
+++ b/pytest_asyncio/plugin.py
@@ -328,18 +328,29 @@ def pytest_report_header(config: Config) -> list[str]:
 
 
 def _fixture_synchronizer(
-    fixturedef: FixtureDef, runner: Runner, request: FixtureRequest
+    fixturedef: FixtureDef,
+    runner: Runner,
+    request: FixtureRequest,
+    runner_fixture_id: str,
 ) -> Callable:
     """Returns a synchronous function evaluating the specified fixture."""
     fixture_function = resolve_fixture_function(fixturedef, request)
     if inspect.isasyncgenfunction(fixturedef.func):
-        return _wrap_asyncgen_fixture(fixture_function, runner, request)  # type: ignore[arg-type]
+        return _wrap_asyncgen_fixture(  # type: ignore[arg-type]
+            fixture_function, runner, request, runner_fixture_id
+        )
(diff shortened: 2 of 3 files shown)


tjkuson (commented):
This doesn't resolve the root issue, which is that sync tests still request the loop factory unparametrized, so the cache key flips and session-scoped fixtures are torn down and rebuilt at every sync/async boundary. The rebuilt fixtures run on the default event loop rather than the configured factory, and sync tests end up with a different fixture instance than async tests in the same session.

Also, CI is failing.

Sanjays2402:
You're right, thanks for spelling out the actual root cause — I was treating the teardown ordering as the problem when it's really a symptom. The unparametrized loop-factory request from sync tests flipping the cache key is the real issue; my patch just reshuffled when the (already wrong) rebuild happens rather than stopping it.

Let me rework this so sync tests request the loop factory the same way async tests do, so the cache key stays stable across the sync/async boundary and session-scoped fixtures aren't torn down and rebuilt on the default loop. I'll fix the CI failures in the same pass. Will push an updated version rather than leave this half-addressed.

tjkuson:
Closing due to inactivity and the needing a lot more work before it's ready. You are welcome to re-open if you ever revisit this PR.

PR #1507, opened 2026-07-15
fix: only warn about asyncio_default_fixture_loop_scope when needed

### LINKED ISSUE #1344: Don't warn about `asyncio_default_fixture_loop_scope` unless the test suite uses `pytest-asyncio`
Hi all. I am not a user of `pytest-asyncio`, but it has gotten installed to my environment through one transitive dependency or another. 

My test suite has no async tests and runs under `-Werror`. One of my tests uses `pytester.runpytest_inprocess`, which fails iff pytest-asyncio is installed due to the deprecation warning:

```
INTERNALERROR>   File "/Users/tybug/.local/share/uv/python/cpython-3.13.11-macos-aarch64-none/lib/python3.13/site-packages/pytest_asyncio/plugin.py", line 247, in pytest_configure
INTERNALERROR>     warnings.warn(PytestDeprecationWarning(_DEFAULT_FIXTURE_LOOP_SCOPE_UNSET))
INTERNALERROR>     ~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
INTERNALERROR> pytest.PytestDeprecationWarning: The configuration option "asyncio_default_fixture_loop_scope" is unset.
INTERNALERROR> The event loop scope for asynchronous fixtures will default to the fixture caching scope. Future versions of pytest-asyncio will default the loop scope for asynchronous fixtures to function scope. Set the default fixture loop scope explicitly in order to avoid unexpected behavior in the future. Valid fixture loop scopes are: "function", "class", "module", "package", "session"
================================================================= short test summary info ==================================================================
FAILED test_a.py::test_a - assert <ExitCode.INTERNAL_ERROR: 3> == 0
==================================================================== 1 failed in 0.06s =====================================================================
 ~/Desktop/pytest_inprocess_repro λ 
```

Reproduction steps:

```bash
mkdir pytest_asyncio_repro && cd pytest_asyncio_repro
uv venv .venv --python 3.12
uv pip install --python .venv/bin/python pytest pytest-asyncio
```

```python                                                                                                                             
# test_a.py
pytest_plugins = ["pytester"]

def test_a(pytester):
    pytester.makepyfile(test_inner="""
def test_b(): pass
""")
    result = pytester.runpytest_inprocess()
    assert result.ret == 0
```

                                                                                                           
* `.venv/bin/python -m pytest test_a.py`: passes
* `.venv/bin/python -m pytest test_a.py -Werror`: fails                                                                                            

I'm not asking for the deprecation warning to be removed or relaxed. But it would be really great if having pytest-asyncio installed in my environment didn't cause my tests to fail!

I also don't know why this requires `pytester.runpytest_inprocess` to trigger. I would expect this to be a problem when running pytest at the top level as well. This may be why nobody has complained about this yet.

pctablet505:
Fixes #1344

When pytest-asyncio is installed as a transitive dependency, the plugin currently emits its `PytestDeprecationWarning` about an unset `asyncio_default_fixture_loop_scope` unconditionally from `pytest_configure`. That means any test suite running under `-Werror` will hit an internal error as soon as pytest-asyncio is present, even if the suite has no async tests or fixtures. The reporter saw this with `pytester.runpytest_inprocess`, but the underlying issue is that the warning fires the moment the plugin loads.

As noted in the issue thread, this happens because pytest enables all installed plugins by default. The suggested workaround of passing `-p no:asyncio` works, but it is not reasonable to expect users who never chose to install pytest-asyncio to discover and apply it.

This change moves the warning into `pytest_fixture_setup`, so it is only emitted when an async fixture is actually being set up and neither the global `asyncio_default_fixture_loop_scope` nor a per-fixture `loop_scope` has been configured. A `StashKey` on the `Config` object keeps track of whether the warning has already been emitted, preventing it from repeating for every async fixture.

I added regression tests for both sides of the behavior: a plain synchronous test now passes under `-Werror` without triggering the warning, while a suite that does use an async fixture still sees the deprecation warning as expected.

### DIFF
--- a/changelog.d/1344.fixed.rst
+++ b/changelog.d/1344.fixed.rst
@@ -1,0 +1,1 @@
+Only warn about an unset ``asyncio_default_fixture_loop_scope`` configuration option when an async fixture is actually used.
--- a/pytest_asyncio/plugin.py
+++ b/pytest_asyncio/plugin.py
@@ -935,12 +935,9 @@ def pytest_fixture_setup(fixturedef: FixtureDef, request) -> object | None:
         if not _is_coroutine_or_asyncgen(fixturedef.func):
             return (yield)
     default_loop_scope = request.config.getini("asyncio_default_fixture_loop_scope")
-    loop_scope = (
-        getattr(fixturedef.func, "_loop_scope", None)
-        or default_loop_scope
-        or fixturedef.scope
-    )
-    if not default_loop_scope and not getattr(fixturedef.func, "_loop_scope", None):
+    fixture_loop_scope = getattr(fixturedef.func, "_loop_scope", None)
+    loop_scope = fixture_loop_scope or default_loop_scope or fixturedef.scope
+    if not default_loop_scope and not fixture_loop_scope:
         _warn_default_fixture_loop_scope_unset(request.config)
     runner_fixture_id = f"_{loop_scope}_scoped_runner"
     runner = request.getfixturevalue(runner_fixture_id)
--- a/pytest_asyncio/plugin.py
+++ b/pytest_asyncio/plugin.py
@@ -293,11 +293,20 @@ def _validate_scope(scope: str | None, option_name: str) -> None:
         )
 
 
+_default_fixture_loop_scope_warning_emitted_for_config: set[int] = set()
+
+
+def _warn_default_fixture_loop_scope_unset(config: Config) -> None:
+    config_id = id(config)
+    if config_id in _default_fixture_loop_scope_warning_emitted_for_config:
+        return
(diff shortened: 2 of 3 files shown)


pctablet505:
Closing this after taking another look at the wider discussion. The comment on #1344 already leaned toward finishing the deprecation rather than making the warning conditional, and that work is underway in #1444 (default the fixture loop scope to function and drop the warning entirely, per #924, targeted at v1.4.0). Once that lands this warning path goes away, so a conditional-suppression workaround here isn't worth review time.

Issue #1227, opened 2025-09-05
23 tests fail: TypeError: fixture() got an unexpected keyword argument 'loop_scope'


yurivict:
[log](https://freebsd.org/~yuri/pytest-asyncio-1.1.0-test.log)

Version: 1.1.0
Python-3.11
FreeBSD 14.3

seifertm:
Thanks for the log. I think I understand what's happening here.

The main pytest call shows that pytest-asyncio-1.1.0 is used:

> plugins: **asyncio-1.1.0**, anyio-4.9.0, aspectlib-2.0.0, xdist-3.6.1, time-machine-2.13.0, typeguard-4.4.4, mpl-0.17.0, cov-5.0.0, hypothesis-6.136.6

However, some of the tests seem to use an older version of pytest-asyncio (v0.23.8):

> running: /usr/local/bin/python3.11 -mpytest --basetemp=/tmp/pytest-of-yuri/pytest-38/test_standalone_test_does_not_trigger_warning_about_no_current_event_loop_being_set0/runpytest-0 --asyncio-mode=strict
     in: /tmp/pytest-of-yuri/pytest-38/test_standalone_test_does_not_trigger_warning_about_no_current_event_loop_being_set0
============================= test session starts ==============================
platform freebsd14 -- Python 3.11.13, pytest-8.1.1, pluggy-1.6.0
Matplotlib: 3.8.0
Freetype: 2.13.3
rootdir: /tmp/pytest-of-yuri/pytest-38/test_standalone_test_does_not_trigger_warning_about_no_current_event_loop_being_set0
configfile: tox.ini
plugins: anyio-4.9.0, aspectlib-2.0.0, xdist-3.6.1, time-machine-2.13.0, **asyncio-0.23.8**, typeguard-4.4.4, mpl-0.17.0, cov-5.0.0, hypothesis-6.136.6
asyncio: mode=Mode.STRICT
collected 1 item


It seems that all affected tests are started in a subprocess using `Pytester.runpytest_subprocess`. I assume that the subprocesses use the system's installation of pytest-asyncio. This would explain why the pytest-asyncio version changes in the middle of the test suite.

However, I don't know why this is the case. It could be an issue with your specific setup or, if subprocess spawning works differently on FreeBSD vs Linux, it could be an issue in pytest.

[...]

PR #1508, opened 2026-07-15
Avoid PytestAssertRewriteWarning in tests using runpytest

### LINKED ISSUE #1334: PytestAssertRewriteWarning due to runpytest and pytest_plugins
_Originally posted by @McSinyx in https://github.com/pytest-dev/pytest-asyncio/issues/1275#issuecomment-3737477316_

On Python 3.11 on Guix ([full test log](https://github.com/user-attachments/files/24579402/python-pytest-asyncio-1.3.0.drv.log)), PytestAssertRewriteWarning resurfaces for the following tests due to 18afc9df5a3153dc1fbdc4e11a56517ef95480df replacing `runpytest` with `runpytest_subprocess`:

    tests/markers/test_class_scope.py::test_standalone_test_does_not_trigger_warning_about_no_current_event_loop_being_set
    tests/markers/test_function_scope.py::test_warns_when_scope_argument_is_present
    tests/markers/test_module_scope.py::test_standalone_test_does_not_trigger_warning_about_no_current_event_loop_being_set
    tests/markers/test_session_scope.py::test_standalone_test_does_not_trigger_warning_about_no_current_event_loop_being_set
    tests/test_event_loop_fixture.py::test_event_loop_fixture_handles_unclosed_async_gen
    tests/test_event_loop_fixture.py::test_event_loop_fixture_asyncgen_error

IIUC this commit tries to workaround a case in GH-1227 where a system `pytest-asyncio` package takes precedence over a local one.  [Our fix for Guix](https://codeberg.org/guix/guix/pulls/4956) also involves removing `pytest_plugins = 'pytest_asyncio'` in the following tests because it also marks the package for assertion rewriting while it's already imported before executing the scripts as part of plugin autodiscovery.

    tests/test_event_loop_fixture.py::test_event_loop_fixture_handles_unclosed_async_gen
    tests/test_event_loop_fixture.py::test_event_loop_fixture_asyncgen_error

pctablet505:
Fixes #1334

Pass `--assert=plain` to the in-process `runpytest` calls for the six tests that emit `PytestAssertRewriteWarning`, keeping the #1227 in-process fix intact. Also drop the explicit `pytest_plugins` assignment in the two event-loop fixture tests so they do not mark the already-loaded plugin for assertion rewriting.

### DIFF
--- a/changelog.d/1334.downstream.rst
+++ b/changelog.d/1334.downstream.rst
@@ -1,0 +1,1 @@
+Avoided a ``PytestAssertRewriteWarning`` in the test suite when pytest-asyncio is already imported, such as when tests are run against a system-installed copy on Guix or FreeBSD.
--- a/changelog.d/1334.downstream.rst
+++ b/changelog.d/1334.downstream.rst
@@ -1,1 +1,1 @@
-Avoided a ``PytestAssertRewriteWarning`` in the test suite when pytest-asyncio is already imported, such as when tests are run against a system-installed copy on Guix or FreeBSD.
+Avoided a ``PytestAssertRewriteWarning`` in the test suite when it is run from an unpacked sdist. The sdist ships ``pytest_asyncio.egg-info/SOURCES.txt``, which lists ``tests/__init__.py``, so pytest marks the ``tests`` package for assertion rewriting even though it has already been imported by the outer session. This affects downstream distributors such as Guix.
--- a/tests/markers/test_class_scope.py
+++ b/tests/markers/test_class_scope.py
@@ -275,5 +275,5 @@ def test_standalone_test_does_not_trigger_warning_about_no_current_event_loop_be
                 async def test_anything(self):
                     pass
             """))
-    result = pytester.runpytest("--asyncio-mode=strict")
+    result = pytester.runpytest("--asyncio-mode=strict", "--assert=plain")
     result.assert_outcomes(warnings=0, passed=1)
--- a/tests/markers/test_class_scope.py
+++ b/tests/markers/test_class_scope.py
@@ -274,5 +274,5 @@ def test_standalone_test_does_not_trigger_warning_about_no_current_event_loop_be
                 async def test_anything(self):
                     pass
             """))
-    result = pytester.runpytest("--asyncio-mode=strict")
+    result = pytester.runpytest("--asyncio-mode=strict", "--assert=plain")
     result.assert_outcomes(warnings=0, passed=1)
--- a/tests/markers/test_function_scope.py
+++ b/tests/markers/test_function_scope.py
@@ -72,7 +72,9 @@ def test_warns_when_scope_argument_is_present(pytester: Pytester):
             async def test_warns():
                 ...
             """))
-    result = pytester.runpytest("--asyncio-mode=strict", "-W", "default")
+    result = pytester.runpytest(
+        "--asyncio-mode=strict", "-W", "default", "--assert=plain"
+    )
     result.assert_outcomes(passed=1, warnings=1)
     result.stdout.fnmatch_lines("*DeprecationWarning*")
 
--- a/tests/markers/test_module_scope.py
+++ b/tests/markers/test_module_scope.py
@@ -281,5 +281,5 @@ def test_standalone_test_does_not_trigger_warning_about_no_current_event_loop_be
             async def test_anything():
                 pass
             """))
-    result = pytester.runpytest("--asyncio-mode=strict")
+    result = pytester.runpytest("--asyncio-mode=strict", "--assert=plain")
     result.assert_outcomes(warnings=0, passed=1)
--- a/tests/markers/test_session_scope.py
+++ b/tests/markers/test_session_scope.py
@@ -392,5 +392,5 @@ def test_standalone_test_does_not_trigger_warning_about_no_current_event_loop_be
             async def test_anything():
                 pass
             """))
-    result = pytester.runpytest("--asyncio-mode=strict")
+    result = pytester.runpytest("--asyncio-mode=strict", "--assert=plain")
     result.assert_outcomes(warnings=0, passed=1)
--- a/tests/test_event_loop_fixture.py
+++ b/tests/test_event_loop_fixture.py
@@ -13,8 +13,6 @@ def test_event_loop_fixture_handles_unclosed_async_gen(
             import asyncio
             import pytest
 
-            pytest_plugins = 'pytest_asyncio'
-
             @pytest.mark.asyncio
             async def test_something():
                 async def generator_fn():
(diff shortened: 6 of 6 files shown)


seifertm (changes requested):
Thanks for the initiative.

The code changes do make sense superficially, but I'm reluctant to integrate them when there's no proof the patch actually solves the problem. Personally, I also don't understand the nature of what causes the issue. In order to progress this, I think we need to find answers to one or more of the following questions:

What is the reason assertion rewriting or setting `pytest_plugins` causes problems in the first place?
Can you provide a reproducer that verifies this patch fixes the linked GUIX issue?
Can you provide a reproducer that verifies for FreeBSD that this patch does not regress #1227 ?

pctablet505:
You were right to hold off — **the patch as I submitted it only fixed 4 of the 6 tests.** Pushed a follow-up. Evidence below.

## Cause

The give-away is in the Guix log: the module pytest names is **`tests`**, not `pytest_asyncio`, and the frame is **`_mark_plugins_for_rewrite`**, not `import_plugin`:

```
_pytest/config/__init__.py:1303: PytestAssertRewriteWarning: Module already imported so cannot be rewritten; tests
    self._mark_plugins_for_rewrite(hook, disable_autoload)
```

So `pytest_plugins` was never the trigger. What happens is that `_mark_plugins_for_rewrite` walks every distribution declaring a `pytest11` entry point and feeds its **file list** to `_iter_rewritable_modules`, which turns any `<dir>/__init__.py` into the bare name `<dir>`. Our wheel is scoped to `packages = ["pytest_asyncio"]`, but the **sdist** ships a pre-generated `pytest_asyncio.egg-info/` whose `SOURCES.txt` lists the whole tree including `tests/__init__.py` — and `importlib.metadata` treats that egg-info as a distribution. Running pytest's own function on the two real metadata files:

```console
$ python -c "
from _pytest.config import _iter_rewritable_modules
lines = [l.strip() for l in open('pytest_asyncio.egg-info/SOURCES.txt') if l.strip()]
print('sdist egg-info SOURCES.txt ->', sorted(set(_iter_rewritable_modules(lines))))
import zipfile
zf = zipfile.ZipFile('dist/pytest_asyncio-1.3.0-py3-none-any.whl')
rec = [l.split(',')[0] for l in zf.read('pytest_asyncio-1.3.0.dist-info/RECORD').decode().splitlines() if l]
print('installed wheel RECORD  ->', sorted(set(_iter_rewritable_modules(rec))))
"
sdist egg-info SOURCES.txt -> ['pytest_asyncio', 'tests']
installed wheel RECORD  -> ['pytest_asyncio']
```

Then `mark_rewrite("tests")` warns because `tests` is already in `sys.modules` under a plain loader — rewriting can't touch an executed module.

**Why CI never sees it:** at config time the build dir isn't on `sys.path` yet, so the egg-info isn't discovered and `tests` isn't marked; collection then puts the build dir on `sys.path` and imports `tests` with an ordinary loader; the in-process `runpytest()` shares `sys.modules`, *now* finds the egg-info, and marks a package that's already imported. `python -m pytest` puts the CWD on `sys.path` at startup, so the first step succeeds and nothing warns — Guix runs the **console script**, which doesn't. That one difference flips it.

## Reproducer

```bash
uv venv venv --python 3.11
uv pip install --python ./venv/bin/python "pytest==9.0.2" hypothesis setuptools \
    setuptools_scm wheel "typing-extensions>=4.12" "backports-asyncio-runner>=1.1,<2"

curl -sLO https://files.pythonhosted.org/packages/source/p/pytest-asyncio/pytest_asyncio-1.3.0.tar.gz
tar xzf pytest_asyncio-1.3.0.tar.gz && cd pytest_asyncio-1.3.0

# build in-tree, like Guix's build phase — this re-reads the shipped SOURCES.txt
../venv/bin/python -c "from setuptools.build_meta import build_wheel; build_wheel('dist')"
uv pip install --python ../venv/bin/python --no-deps dist/*.whl

../venv/bin/pytest          #  -> 6 failed, 234 passed
../venv/bin/python -m pytest #  -> 240 passed
```

`6 failed, 234 passed` — the same 6 tests as the Guix log, out of the same 240, with the same warning text and frame. Same tree under `-m` passes.

## Which half of the patch does the work

Five variants, clean restore of the pristine sdist each time, full suite:

| variant | result |
|---|---|
| unpatched | `6 failed, 234 passed` |
| **this PR as submitted** | **`2 failed, 238 passed`** |
| `--assert=plain` on all 6, keep `pytest_plugins` | `240 passed` |
| `--assert=plain` on all 6, drop `pytest_plugins` | `240 passed` |
| only drop `pytest_plugins` | `6 failed, 234 passed` |

`--assert=plain` is the whole fix and it needs to be on all six — I'd only put it on four. Dropping `pytest_plugins = 'pytest_asyncio'` does nothing either way, since the plugin is autoloaded through its entry point; that's also the answer to @McSinyx's question about what removing it does. **Happy to drop that hunk** if you'd rather keep the diff to `--assert=plain` only.

Worth noting this isn't a new idiom: `18afc9d` — the commit that fixed #1227 — already added `--assert=plain` while converting `runpytest_subprocess` to `runpytest`, 17 times across `test_asyncio_mark.py`, `modes/test_strict_mode.py`, `markers/test_invalid_arguments.py`, `test_fixture_loop_scopes.py` and `test_event_loop_fixture.py`. It just didn't apply it to every test it converted — `test_event_loop_fixture.py` got it on one of its three tests and not the other two, which are two of the six failing here. This PR finishes that.

## On #1227

That was a different failure — `TypeError: fixture() got an unexpected keyword argument 'loop_scope'`, from `runpytest_subprocess` picking up the system's 0.23.8 — so it's purely about which pytest-asyncio the inner run imports. This PR doesn't touch that axis; the call inventory over `tests/` is identical either side, in every file:

```
main: {'runpytest': 155, 'runpytest_subprocess': 6}
PR  : {'runpytest': 155, 'runpytest_subprocess': 6}
```

Checked empirically too, with 0.23.8 in a directory a subprocess finds on `PYTHONPATH` while the outer session holds the correct version — three `pytester` runs printing the inner plugin banner:

```
IN-PROCESS  (default)      : plugins: asyncio-1.3.0, hypothesis-6.161.6
IN-PROCESS  (--assert=plain): plugins: asyncio-1.3.0, hypothesis-6.161.6
SUBPROCESS  (default)      : plugins: asyncio-0.23.8, hypothesis-6.161.6
```

The subprocess reproduces #1227 on demand; `--assert=plain` leaves the in-process banner byte-identical, since it only controls whether the rewrite hook is installed, not import resolution. No FreeBSD hardware here, so that's the mechanism plus a Linux reproduction of the same shadowing, not a literal FreeBSD run.

## Pushed

`--assert=plain` added to the two remaining tests in `tests/test_event_loop_fixture.py`, and a corrected changelog fragment (the old one blamed a system-installed copy, which isn't the mechanism). Verified against an sdist built from this branch: **`2 failed, 296 passed` → `298 passed`**.

One thing for you to decide: the more principled fix is probably packaging-side — not shipping `tests/` paths in the sdist's `SOURCES.txt`, or not making the suite an importable `tests` package — which would cover every downstream distributor and any future `runpytest()` test at once. Bigger change and your call, so I kept this minimal; happy to take a run at it if you'd prefer.

PR #1517, opened 2026-07-21
Remove deprecated event_loop_policy fixture

seifertm:
The asyncio policy system has been deprecated and is scheduled for removal in CPython 3.16. The pytest-asyncio policy system is deprecated and has been superseeded by the *pytest_asyncio_loop_factory* hook.

This PR removes the deprecated *event_loop_policy* fixture.

Obsoletes #796


tjkuson (approved):

pytest_asyncio/plugin.py
```diff
@@ -805,11 +805,8 @@ def _temporary_event_loop(loop: AbstractEventLoop) -> Iterator[None]:
 
 
 @contextlib.contextmanager
-def _temporary_event_loop_policy(
-    policy: AbstractEventLoopPolicy,
-) -> Iterator[None]:
+def _restore_event_loop_policy() -> Iterator[None]:
```
edgarrmondragon, line 808: Should this context manager be outright removed? It still calls deprecated/removed APIs and all tests under Python 3.14 and 3.16 pass when I don't use it in `_create_scoped_runner_fixture`.

tjkuson: We discussed this offline at the pytest sprint. I think there are merits to either approach.

The idea was that code under test might apply an event loop policy which will need to be restored between tests even if we remove support for configuring the policy via the pytest-asyncio fixture. If we remove policy restoration context manager, users who depend on event loop policy being reset between tests could see confusing errors when upgrading to this new version (independent of the deprecated loop policy fixture being removed).

However, getting a warning about the deprecated API even after we ostensibly remove support for it is also a bad user experience. It might be strong enough reason to remove the support for event loop policies entirely.

If we remove both the fixture and the feature of the plugin that reset the policy between tests, then we'd need to communicate to users that both the fixture and support for testing code that sets an event loop policy are unsupported in v2.

edgarrmondragon: Thanks for the context! So perhaps there is a compromise where the current policy is guarded and restored _only_ on Python < 3.16, since users won't presumably be able to apply any policies on 3.16+ anyway. But perhaps that's thinking way too ahead of time.

PR #1519, opened 2026-07-22
docs(how-to): Add guide for isolating async applications in tests

AreteDriver:
## Summary

Adds a new how-to guide demonstrating two common failure modes when testing async applications (e.g., FastAPI, Quart, aiohttp) with pytest-asyncio, and how to fix them using fixture-scoped imports and temporary directory overrides.

## Motivation

Testing async web frameworks often involves importing an `app` object and overriding dependencies. Two subtle isolation traps repeatedly bite production codebases:

1. **Stale app reference via module-level import** — `importlib.reload` creates a new app instance, but other test modules hold stale references to the old one through module-level imports.
2. **Default path leak** — An app or service with a default file-system path in its constructor writes state that bleeds across tests.

pytest-asyncio's fixture scoping makes these easy to prevent, but the failure modes aren't well documented anywhere in the asyncio ecosystem.

## Scope

- `docs/how-to-guides/isolate_async_apps.rst` — the guide
- `docs/how-to-guides/isolate_async_apps_example.py` — runnable examples
- `docs/how-to-guides/index.rst` — add entry to toctree

## Verification

The example code passes on pytest-asyncio 0.24+ with FastAPI as the demo framework (any async web framework exhibits the same pattern). Both failure modes are reproduced with failing assertions, then fixed.

---

**Author's note:** These patterns were extracted from production test suites (github.com/AreteDriver/animus and github.com/AreteDriver/BenchGoblins). Happy to adjust to match project conventions.

### DIFF
--- a/docs/how-to-guides/index.rst
+++ b/docs/how-to-guides/index.rst
@@ -20,5 +20,6 @@ How-To Guides
   parametrize_with_asyncio
   uvloop
   test_item_is_async
+  isolate_async_apps
 
 This section of the documentation provides code snippets and recipes to accomplish specific tasks with pytest-asyncio.
--- a/docs/how-to-guides/isolate_async_apps.rst
+++ b/docs/how-to-guides/isolate_async_apps.rst
@@ -1,0 +1,60 @@
+===============================================
+How to isolate async applications in your tests
+===============================================
+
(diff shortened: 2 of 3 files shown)


tjkuson:
Thank you for your interest in improving the documentation. However, I do not think this how-to guide belongs as part of pytest-asyncio (it's guidance about shared state rather than anything specific to the plugin). I also think the advice in too opinionated to merge (for starters, I wouldn't recommend `importlib.reload` as a solution to this problem).

PR #1528, opened 2026-07-23
Simplify testing minimum direct dependencies

tjkuson:
Instead of defining a pytest-min dependency group to test against the lowest supported dependencies (which needs to be maintained), configure tox to use uv's lowest-direct resolution strategy to achieve the same thing.

Rename the test env to be more accurate (it's not just pytest minimum dependencies it checks). Prefixing with `py310` means `tox -e py310-lower-bounds` will use Python 3.10 instead of relying on the `gh-actions` configuration.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -82,21 +82,6 @@ build = [
   "check-wheel-contents",
   "twine",
 ]
-pytest-min = [
-  "attrs==23.1",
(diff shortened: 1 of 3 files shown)


seifertm (approved):
Very nice! This also avoids having to maintain the minimum supported pytest version in multiple places.

PR #1548, opened 2026-08-20
Add cooperative signal timeouts

tamird:
A signal timeout can interrupt asyncio while it schedules a task's next step, leaving a pending task with no continuation [1]. Catching the resulting failure and cancelling that task cannot reliably recover it.

Integrate automatically on Python 3.11 and newer when pytest-timeout provides the proposed expiry hook [2]. No configuration is needed. The signal handler records the supplied failure and queues cancellation without raising into the scheduler. asyncio.timeout owns cancellation accounting; the runner reports the original timeout with the interrupted await's traceback while preserving process-control exceptions and external cancellation. A timeout during final runner shutdown keeps stopping the loop across shutdown phases until Runner.close() exits; remaining resource cleanup may be incomplete.

Python 3.10, pytest-timeout versions without the hook, and synchronous test coroutine creators retain their existing signal behavior. Cooperative cancellation cannot interrupt CPU-bound code that never yields, indefinitely blocking callbacks, or tasks that refuse cancellation. pytest-timeout's thread method remains the hard process-stop option. This consumer does not take over other async plugins or synchronous tests that call asyncio.run().

This keeps runner integration separate from timer policy, as requested in the earlier timeout discussion [3]. The dedicated integration environment pins the producer prerequisite; that pin should become a released test dependency once the hook is available. Related to #215.

Prepared with Codex assistance.

[1]: https://github.com/pytest-dev/pytest-timeout/issues/113
[2]: https://github.com/pytest-dev/pytest-timeout/pull/204
[3]: https://github.com/pytest-dev/pytest-asyncio/pull/216#issuecomment-1578935413

### DIFF
--- a/docs/reference/timeouts.rst
+++ b/docs/reference/timeouts.rst
@@ -1,0 +1,28 @@
+========
+Timeouts
+========
+
+Enable ``--asyncio-cooperative-timeouts`` or set
+``asyncio_cooperative_timeouts = true`` to deliver pytest-timeout signal
+failures by cancelling the active asynchronous test or fixture. This requires
+a pytest-timeout version providing ``pytest_timeout_expired``. Cooperative
+cleanup can then run before pytest reports the original timeout.
+pytest-timeout still controls the configured duration, covered test phases,
+debugger detection, and timeout diagnostics.
+
+Cancellation waits for the event loop and coroutine to cooperate. It cannot
+stop a callback that blocks indefinitely or a task that refuses cancellation.
+Use pytest-timeout's ``thread`` method or an independent process watchdog when
+the process must be terminated. A timeout during final event-loop shutdown
+stops that shutdown and closes the loop; remaining resource cleanup may be
+incomplete.
+
+Custom Python 3.10 task factories may return tasks without cancellation
+counters. For those tasks, an escaping ``CancelledError`` is preserved and
+the timeout is chained to it, because the runner cannot reliably distinguish
+timeout cancellation from another cancellation request.
+
+Cooperative timeouts are disabled by default. Enabling them without a
+compatible pytest-timeout plugin is a configuration error. The integration
+does not take over runners managed by other async plugins or synchronous tests
+that call ``asyncio.run()`` themselves.
--- a/docs/reference/timeouts.rst
+++ b/docs/reference/timeouts.rst
@@ -2,27 +2,25 @@
 Timeouts
 ========
 
-Enable ``--asyncio-cooperative-timeouts`` or set
-``asyncio_cooperative_timeouts = true`` to deliver pytest-timeout signal
-failures by cancelling the active asynchronous test or fixture. This requires
-a pytest-timeout version providing ``pytest_timeout_expired``. Cooperative
-cleanup can then run before pytest reports the original timeout.
+On Python 3.11 and newer, pytest-asyncio automatically delivers pytest-timeout
+signal failures by cancelling the active asynchronous test or fixture when
+pytest-timeout provides the ``pytest_timeout_expired`` hook. Cooperative cleanup
+can then run before pytest reports the original timeout. No configuration is
+needed. Python 3.10 and older pytest-timeout versions retain their existing
+signal behavior.
 pytest-timeout still controls the configured duration, covered test phases,
 debugger detection, and timeout diagnostics.
 
-Cancellation waits for the event loop and coroutine to cooperate. It cannot
-stop a callback that blocks indefinitely or a task that refuses cancellation.
+Cancellation waits for the event loop and coroutine to cooperate. A raised
+signal exception can interrupt CPU-bound Python code, but cooperative
+cancellation cannot interrupt code that never yields. It also cannot stop
+an indefinitely blocking callback or a task that refuses cancellation.
 Use pytest-timeout's ``thread`` method or an independent process watchdog when
-the process must be terminated. A timeout during final event-loop shutdown
-stops that shutdown and closes the loop; remaining resource cleanup may be
-incomplete.
+the process must be terminated; these stop the entire process without normal
+test teardown. A timeout during final event-loop shutdown stops that shutdown
+and closes the loop; remaining resource cleanup may be incomplete.
 
-Custom Python 3.10 task factories may return tasks without cancellation
-counters. For those tasks, an escaping ``CancelledError`` is preserved and
-the timeout is chained to it, because the runner cannot reliably distinguish
-timeout cancellation from another cancellation request.
-
-Cooperative timeouts are disabled by default. Enabling them without a
-compatible pytest-timeout plugin is a configuration error. The integration
-does not take over runners managed by other async plugins or synchronous tests
-that call ``asyncio.run()`` themselves.
+The integration does not take over runners managed by other async plugins or
+synchronous tests that call ``asyncio.run()`` themselves. It wraps asynchronous
+tests and fixtures in a timeout context, so ``asyncio.current_task().get_coro()``
+returns the wrapper coroutine rather than the original test coroutine.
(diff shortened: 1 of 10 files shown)


docs/reference/timeouts.rst
```diff
@@ -0,0 +1,28 @@
...
+the process must be terminated. A timeout during final event-loop shutdown
+stops that shutdown and closes the loop; remaining resource cleanup may be
+incomplete.
+
+Custom Python 3.10 task factories may return tasks without cancellation
+counters. For those tasks, an escaping ``CancelledError`` is preserved and
+the timeout is chained to it, because the runner cannot reliably distinguish
+timeout cancellation from another cancellation request.
+
+Cooperative timeouts are disabled by default. Enabling them without a
```
tjkuson: I am curious, what was the motivation for this to be disabled by default and enabled via a configuration? What would be the reason for a user of both pytest-asyncio and pytest-timeout not enable this feature?

Perhaps I am mistaken, but I see this PR more as a compatibility layer for a (possible) pytest-timeout hook rather than something a user would want to configure. As a maintainer, I'd rather say "this plugin integrates with pytest-timeout" instead of "you need to enable these configurations to get these plugins to work together."

pytest_asyncio/_timeout.py
```diff
@@ -0,0 +1,175 @@
...
+    invocation = _Invocation(runner.get_loop())
+
+    async def invoke() -> _T:
+        task = asyncio.current_task()
+        assert task is not None
+        invocation.task = task
+        if invocation.exception is not None:
+            coro.close()
+            raise invocation.exception
+        # A custom Python 3.10 task factory can bypass the runner backport's
```
tjkuson: This seems like quite a lot of work to support 3.10 which is nearly EOL and does not need to be supported by pytest-asyncio for much longer. The PR that removes 3.10 support should probably be its own thing given we have 3.10-specific code already in the codebase, but for the purpose of this PR I am in favour of just treating 3.10 as unsupported. It should simplify things.

Likewise, we could use `asyncio.timeout` and `asyncio.Timeout`.

pytest_asyncio/_timeout.py
```diff
@@ -0,0 +1,175 @@
...
+    *,
+    context: contextvars.Context,
+    config: pytest.Config,
+) -> _T:
+    if _RUNNER_STATE not in config.stash:
+        return runner.run(coro, context=context)
+
+    invocation = _Invocation(runner.get_loop())
+
+    async def invoke() -> _T:
```
tjkuson, line 109: This wrapper changes the test coroutine which messes with the traceback and changes `task.get_coro`. The first could probably be solved by setting `__tracebackhide__`, but I am not sure if anything can be done about the second issue (though I am also not convinced that it's blocking given we already wrap fixtures).

pytest_asyncio/_timeout.py
```diff
@@ -0,0 +1,175 @@
...
+
+
+@dataclass
+class _Shutdown(_Delivery):
+    def interrupt(self, state: _RunnerState) -> None:
+        self.handle = None
+        if state.invocation is self:
+            # Runner.close() owns the loop and closes it in a finally block.
+            # Stop only that final shutdown, never a reusable runner invocation.
+            self.loop.stop()
```
tjkuson, line 59: `Runner.close()` runs several blocking phases, each via `run_until_complete`. If the timeout lands as one phase completes, that phase consumes the stop and `close()` can proceed into a blocking `shutdown_default_executor()` after the only alarm has fired and hang indefinitely. We might need some way to persist the stop...

tjkuson (changes requested):
Hi, thanks for the PR! Improving how timeouts for users of pytest-asyncio and pytest-timeout would be great, and this new hook is an interesting proposal.

I've left some comments from a pytest-asyncio angle. I'm not directly involved with the pytest-timeout plugin and can't speak on their behalf regarding whether this new hook works for them.

docs/reference/timeouts.rst
tamird: The opt-in was intended to preserve interruption semantics: raising from `SIGALRM` can interrupt CPU-bound Python code, whereas cooperative cancellation needs the event loop to run. I agree that requiring another configuration switch is undesirable.

I've now made the integration automatic on Python 3.11+ when the expiry hook is available, and the documentation makes that tradeoff explicit. Users needing hard termination can use the thread method, which terminates the whole process rather than continuing the test run.

pytest_asyncio/_timeout.py
tamird: Agreed. The integration now uses `asyncio.timeout(None)` on Python 3.11+, with expiry rescheduled from the queued event-loop callback. That lets the standard library own the cancellation accounting and removes the custom Python 3.10 task fallback. Python 3.10 retains the existing signal behavior; the project's supported Python versions are unchanged.

pytest_asyncio/_timeout.py
tamird: Added `__tracebackhide__` to the internal delivery frames and checked that the interrupted application frame remains visible. The `get_coro()` difference remains: the timeout context has to be entered inside the task executing the test, so this design still needs the wrapper. That behavior is now documented. I'd prefer to keep the wrapper rather than take over task creation or introduce another task just to preserve coroutine introspection.

pytest_asyncio/_timeout.py
tamird: Fixed in b96aff4. The stop callback now requeues itself until `Runner.close()` exits, so completing one shutdown phase cannot consume the only stop and leave the next phase blocked. The remaining callback is canceled on exit. The regression triggers expiry at the async-generator/executor shutdown boundary with an executor job still blocked; it passes with this fix and fails with the previous one-shot stop.

tamird (commented):
full disclosure: i'm doing by best but the intricacies of asyncio are beyond the depth of my knowledge so i'm leaning on codex pretty hard.

if at any point you feel like you're better off picking this up do let me know, I won't be offended. I still haven't heard anything on the pytest-timeout of this proposal.

Thanks for reviewing!

pytest_asyncio/_timeout.py
tamird: Follow-up on the callback detail: on [the current head](https://github.com/pytest-dev/pytest-asyncio/blob/f729f0f57dbc1619903c065dade220b48c2e756e/pytest_asyncio/_timeout.py#L55-L62), the stop still requeues across shutdown phases until `Runner.close()` exits. There is no saved handle to cancel now: callbacks check that their delivery is still active and otherwise return. This avoids relying on receiving a handle before SIGINT interrupts scheduling. The async-generator/executor boundary regression still passes locally.

Posted with Codex assistance.

pytest_asyncio/_timeout.py
tjkuson: I think the `get_coro` difference is acceptable (and I can't really see a good way around it).

pytest_asyncio/_timeout.py
tjkuson: Thanks, this addresses the race I was concerned about, though I will leave this thread open given it's probably the most nuanced thing in this PR and will think about it some more.

tjkuson:
Thank you for the substantial follow-up work!

I've reviewed the changes and consider my earlier comments addressed. I don't currently see any correctness issues with the approach. However, we can't merge this unless https://github.com/pytest-dev/pytest-timeout/pull/204 is accepted and released (which is out of my control). If or when that happens, I'll do another pass and maybe make some adjustments, though nothing I've seen so far has my questioning the overall design. I'd also like to leave room for other reviewers.

> full disclosure: i'm doing by best but the intricacies of asyncio are beyond the depth of my knowledge so i'm leaning on codex pretty hard.

I appreciate the transparency!

PR #1551, opened 2026-08-28
Apply loop factories when asyncio is marked via parametrize

### LINKED ISSUE #1463: `pytest_asyncio_loop_factories` hook is not called when `pytest.mark.asyncio` is added via parametrization

gyanu2507:
`pytest_generate_tests` only looked at the function's own asyncio marker, so this never picked up loop factories:

```python
@pytest.mark.parametrize(
    "backend",
    [pytest.param("asyncio", marks=pytest.mark.asyncio)],
)
async def test_async(backend):
    ...
```

Also convert those functions during collection so strict mode still owns them.

Fixes #1463.


tjkuson (changes requested):
Thank you for the PR, but this doesn't seem to address the issues I raised under the original issue. Specifically, this seems to break code which mixes asyncio and trio markers. https://github.com/pytest-dev/pytest-asyncio/issues/1463#issuecomment-4595539867

gyanu2507:
You're right — this also wraps the trio param set. Closing; a proper fix needs the pytest API you pointed at.

Issue #1514, opened 2026-07-21
Warn when a fixture for a different loop_scope is requested


seifertm:
Pytest-asyncio currently allows requesting fixtures that have a different *loop_scope* than the requesting test or fixture. Since fixtures with different loop scopes use different asyncio event loops, this can lead to unexpected errors., e.g. trying to access a task that was created on a different loop. 

This can happen when: 
* a test requests a fixture, but the *loop_scope*s don't match
* a fixture requests another fixture, but the *loop_scope*s don't match

Mixing fixtures with different loop scopes is almost certainly a user error and should result in a warning.

*Work on this issue was done during the pytest sprint 2026, sponsored by [Omicron](https://www.omicronenergy.com/en/).*

tjkuson:
Emitting a warning when pytest 'jumps' a loop scope during the execution of a test and its fixtures makes a lot of sense to me. However, users of pytest-asyncio currently can configure the default loop scope for fixtures and tests separately. I think it also follows that it is almost certain that configuring the default loop scopes for fixtures and tests to be different is likely a user error as well.

To avoid user confusion, I think it would make sense to consolidate these configs to a single `asyncio_default_loop_scope` that controls the default loop scope for both fixtures and tests. We could then deprecate the current configurations and point people to use the new one. The way this would be communicated to users could be along the lines of:

> pytest-asyncio defaults to running each test function and any async fixture it requests in its own function-scoped event loop. If most of your async tests use async fixtures that use a wider loop scope (e.g., a session-scoped async fixture that requires a session-scoped event loop), you should configure the default loop scope accordingly. You can control the loop scope on individual test functions and fixtures using the `pytest.mark.asyncio` marker and `pytest_asyncio.fixture` decorator.

PR #1552, opened 2026-08-29
Warn when fixture loop scopes differ

sunilkumarvalmiki:
## Summary

- warn when an async fixture is requested by a test or async fixture using a different event-loop scope
- add regression coverage for test-to-fixture and fixture-to-fixture mismatches
- add the changelog entry and keep the subprocess warning check scoped to `ResourceWarning`

## Tests

- `.venv\\Scripts\\ruff.exe check pytest_asyncio/plugin.py tests/test_fixture_loop_scopes.py tests/test_set_event_loop.py`
- `.venv\\Scripts\\python.exe -m pytest tests/test_fixture_loop_scopes.py -q` (13 passed)
- `.venv\\Scripts\\python.exe -m pytest -q` (297 passed, 3 skipped)

Refs #1514


pytest_asyncio/plugin.py
```diff
@@ -342,6 +342,57 @@ def _fixture_synchronizer(
...
+    ):
+        return (
+            f"fixture {parent_fixturedef.argname!r}",
+            _get_fixture_loop_scope(parent_fixturedef, request.config),
+        )
+
+    marker = request.node.get_closest_marker("asyncio")
+    if marker is None:
+        return None
+    return (
```
tjkuson, line 369: For fixtures with a non-function scope, `request.node` refers to the scope's collector, not the test requesting the fixture. Thus, mismatches with the test’s loop scope are missed.

pytest_asyncio/plugin.py
```diff
@@ -925,12 +976,8 @@ def pytest_fixture_setup(fixturedef: FixtureDef, request) -> object | None:
...
         if not _is_coroutine_or_asyncgen(fixturedef.func):
             return (yield)
-    default_loop_scope = request.config.getini("asyncio_default_fixture_loop_scope")
-    loop_scope = (
-        getattr(fixturedef.func, "_loop_scope", None)
-        or default_loop_scope
-        or fixturedef.scope
-    )
+    loop_scope = _get_fixture_loop_scope(fixturedef, request.config)
+    _warn_if_fixture_loop_scope_mismatch(fixturedef, request, loop_scope)
```
tjkuson, line 980: `pytest_fixture_setup` only runs when the fixture is first created, so later requests bypass this check. The check shouldn't depend on test order/trigger inconsistently.

pytest_asyncio/plugin.py
```diff
@@ -342,6 +342,57 @@ def _fixture_synchronizer(
...
+        or fixturedef.scope
+    )
+
+
+def _get_requesting_loop_scope(
+    request: FixtureRequest,
+) -> tuple[str, _ScopeName] | None:
+    parent_request = getattr(request, "_parent_request", None)
+    parent_fixturedef = getattr(parent_request, "_fixturedef", None)
+    if parent_fixturedef is not None and _is_asyncio_fixture_function(
```
tjkuson, line 358: As written, this doesn't work with pytest-asyncio in auto mode. The warning should work on both auto and non-auto modes.

tjkuson (changes requested):
Hi, thanks for the PR. I left some comments regarding the warning being skipped depending on how fixtures are requested, which is a problem as the purpose of the warning is to catch these subtle mismatches. It also reaches into pytest internals which we would ideally avoid.

PR #1553, opened 2026-08-31
Fix loop factory parametrization for sync fixtures

### LINKED ISSUE #1501: Loop factory parametrization tears down async fixtures used by sync tests

TLNing260310:
Fixes #1501.

## Summary

- Parametrize synchronous tests over configured loop factories when their static fixture graph contains a pytest-asyncio-managed fixture.
- Run the plugin's generation hook after pytest and user parametrization hooks so direct parameters correctly shadow same-named fixtures while indirect parameters keep using their fixture definitions.
- Use the widest effective managed-fixture loop scope and resolve the loop-factory parameter before dependent fixtures inspect their caches.
- Cover fixture overrides, transitive dependencies, autouse fixtures, auto mode, teardown order, and direct/indirect parametrization; document the static-closure boundary.

## Why

Previously only async test items requested the parametrized loop-factory fixture. A synchronous test consuming a managed async fixture therefore requested an unparametrized/default factory, changed the factory cache key, and could tear down shared fixtures at a sync/async test boundary. Parametrizing the affected synchronous test items keeps the factory cache key and fixture lifetime consistent.

## Validation

- `tests/test_loop_factory_parametrization.py`: 55 passed with pytest 8.4.0
- `tests/test_loop_factory_parametrization.py`: 55 passed with pytest 9.1.1
- Broader pytest 8.4 run: 306 passed, 1 unrelated Windows nested-pytest test deselected; the local 8.4 environment did not include Hypothesis
- Broader pytest 9.1 run: 308 passed, 3 skipped, 1 unrelated Windows nested-pytest test deselected
- `ruff check pytest_asyncio tests`
- `pyright pytest_asyncio tests`

The documentation build was not run locally because Sphinx was not installed in the test environment.

### DIFF
--- a/pytest_asyncio/plugin.py
+++ b/pytest_asyncio/plugin.py
@@ -691,47 +691,15 @@ def _get_managed_fixture_loop_scope(
     """Return the widest loop scope of managed fixtures used by a sync test."""
     asyncio_mode = _get_asyncio_mode(metafunc.config)
     default_loop_scope = metafunc.config.getini("asyncio_default_fixture_loop_scope")
-    fixtureinfo = metafunc.definition._fixtureinfo
-    unresolved_initialnames = {
-        fixture_name
-        for fixture_name in fixtureinfo.initialnames
-        if fixture_name not in metafunc._arg2fixturedefs
-    }
-    discovered_fixturedefs: dict[str, Sequence[FixtureDef] | None] = {}
-
-    def get_fixturedefs(fixture_name: str) -> Sequence[FixtureDef] | None:
-        fixturedefs = metafunc._arg2fixturedefs.get(fixture_name)
-        if fixturedefs is not None:
-            return fixturedefs
-        # Directly parametrized test arguments have no static FixtureDef and must
-        # not fall back to a fixture with the same name.
-        if fixture_name in unresolved_initialnames:
-            return None
-        if fixture_name not in discovered_fixturedefs:
-            # pytest 8.4 does not add dependencies of overridden super fixtures
-            # to names_closure, so resolve those dependencies on demand.
-            discovered_fixturedefs[fixture_name] = (
-                metafunc.definition.session._fixturemanager.getfixturedefs(
-                    fixture_name, metafunc.definition
-                )
-            )
-        return discovered_fixturedefs[fixture_name]
-
     loop_scopes: list[Scope] = []
-    current_fixturedef_indices: dict[str, int] = {}
-
-    def collect_loop_scopes(fixture_name: str) -> None:
-        fixturedef_index = current_fixturedef_indices.get(fixture_name)
-        if fixturedef_index == -1:
-            return
-        fixturedefs = get_fixturedefs(fixture_name)
+    # Let pytest own fixture traversal and shadowing. There is no public API for
+    # retrieving the active FixtureDef from Metafunc, so this is intentionally a
+    # shallow, read-only use of pytest's already-computed fixture closure.
+    for fixture_name in metafunc.fixturenames:
+        fixturedefs = metafunc._arg2fixturedefs.get(fixture_name)
         if not fixturedefs:
-            return
-        if fixturedef_index is None:
-            fixturedef_index = -1
-        if -fixturedef_index > len(fixturedefs):
-            return
-        fixturedef = fixturedefs[fixturedef_index]
+            continue
+        fixturedef = fixturedefs[-1]
         func = fixturedef.func
         is_managed = _is_asyncio_fixture_function(func) or (
             asyncio_mode == Mode.AUTO and _is_coroutine_or_asyncgen(func)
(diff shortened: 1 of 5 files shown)


pytest_asyncio/plugin.py
```diff
@@ -774,7 +855,21 @@ def pytest_generate_tests(metafunc: pytest.Metafunc) -> None:
...
             for name in marker_selected_factory_names
         ]
-    metafunc.fixturenames.append(_asyncio_loop_factory.__name__)
+    if sync_test_uses_managed_fixture:
+        # Resolve the parameter before any managed fixture checks its cache. This
+        # ensures a factory change tears down and rebuilds wider-scoped fixtures
+        # before their values are supplied to a synchronous test.
+        fixtureinfo = metafunc.definition._fixtureinfo
+        if _asyncio_loop_factory.__name__ not in fixtureinfo.initialnames:
+            object.__setattr__(
```
tjkuson: We should not be manipulating pytest internals like this.

pytest_asyncio/plugin.py
```diff
@@ -724,20 +799,26 @@ def pytest_pycollect_makeitem_convert_async_functions_to_subclass(
     hook_result.force_result(updated_node_collection)
 
 
-@pytest.hookimpl(tryfirst=True)
+# Direct parametrization replaces same-named fixtures during this hook. Run after
+# pytest and user hooks so the resolved fixture graph reflects those replacements.
+@pytest.hookimpl(trylast=True)
```
tjkuson: This changes the node IDs which is breaking for users that match on test names.

pytest_asyncio/plugin.py
```diff
@@ -685,6 +685,81 @@ def _resolve_asyncio_marker(item: Function) -> Mark | None:
     return None
 
 
+def _get_managed_fixture_loop_scope(
```
tjkuson, line 688: This reimplements pytest's fixture traversal system in a way that's error-prone is difficult to maintain.

tjkuson (changes requested):
Thank you for the PR, but as-written it duplicates, coerces, and undermines pytest's internals in a way that's error-prone and difficult to maintain.

pytest_asyncio/plugin.py
TLNing260310: Addressed in d86d02c. I removed the object.__setattr__ call and no longer mutate FuncFixtureInfo.initialnames. After pytest has created and pruned the Function item, the existing collection wrapper places the already-parametrized internal factory fixture first in the item's fixture resolution order, so its cache key changes before managed fixtures are read.

pytest_asyncio/plugin.py
TLNing260310: Addressed in d86d02c. The recursive fixture traversal, FixtureManager fallback, and override-chain bookkeeping have been removed. Sync detection now only scans the pytest-computed metafunc.fixturenames closure and reads the active fixturedefs[-1]. The docs explicitly limit the behavior to dependencies represented in the static closure rather than reimplementing traversal for dynamic or same-name override cases.

pytest_asyncio/plugin.py
TLNing260310: Addressed in d86d02c. The original async pytest_generate_tests(tryfirst=True) ordering is restored. Sync fixture detection now uses a separate post-yield wrapper, so existing async IDs retain the factory-user_parameter order. I added collect-only regression coverage for both the preserved async order and the new sync user_parameter-factory order.

TLNing260310:
Thanks for the review. I reworked the implementation in d86d02c around the three concerns: the custom recursive fixture traversal and FixtureManager fallback are gone; FuncFixtureInfo.initialnames is no longer mutated; and the original tryfirst ordering for async tests is restored with collect-only node-ID regression tests. Sync tests now use a separate post-yield generate wrapper, scan only the pytest-computed static fixture closure, and resolve the factory cache key before managed fixtures. The documented boundary is intentionally the closure pytest exposes: dynamic request.getfixturevalue dependencies and hidden same-name override chains are not rediscovered. The only remaining private access is a shallow, read-only lookup of the active FixtureDef through Metafunc._arg2fixturedefs because pytest 8.4-9.1 has no public equivalent. All CI checks pass, including pytest main, Python 3.10-3.14/3.14t on Windows and Linux, docs, lint, and packaging. Codecov reports all modified coverable lines covered. I would appreciate another review.

PR #1556, opened 2026-09-05
docs: add AI policy

tjkuson:
Having a policy to set expectations around AI-assistance in contributions will likely still be ignored by the worst actors but will at a minimum give us something to reference. It's deliberately succinct in order to be read and understood quickly.


Pierre-Sassoulas (commented):
Might want to just refer to pytest's AI policy ?

tjkuson:
> Might want to just refer to pytest's AI policy ?

I did consider that, but the pytest AI policy is quite long and repository-specific. I mentioned it briefly at the pytest sprint a few weeks ago where it seemed like the intention was that projects under pytest-dev would write project-specific AI policies at their own discretion. FWIW, I think an org-wide policy would have benefits, but there'd probably need to be some sort of coordination mechanism between all the projects. If we pointed potential pytest-asyncio contributors to the pytest policy, we'd both need to watch for changes to that upstream policy, and it'd be a more work to edit it.

For now, I personally prefer that pytest-asyncio adopts this smaller AI-policy (that's effectively a simplified version of pytest's).

seifertm (approved):
I agree that we eventually might want to converge on pytest's policy, but this is a good start. Thanks, Tom!

PR #1557, opened 2026-09-07
Use issue_config_time_warning for the unset-loop-scope deprecation

### LINKED ISSUE #1142: Ambiguous event loop overrides
Hi. First of all, thanks for the library — it is really a timesaver! ;-)

Now, to the point: I maintain a fancy library [looptime](https://github.com/nolar/looptime) that compacts the loop time into zero true (aka wall-clock) time, while simulating the time flow within the loop (be that seconds or years of loop time in 0.1s of true time). Previously (pytest-asyncio<1.0.0), every test had its own event loop and lived in its own universe from the event loop creation (say, time 0). Now (>=1.0.0), the event loop can be shared across many tests at different scopes, also separated by modules/classes/packages. I am trying to adapt my library to this behaviour.

I have found a somewhat confusing case which is easy to hit — see a repro here.

The key point is that a regular fixture `fixt` is function-scoped by default, and is intended to be so. The rest of the code is basically an example from pytest-asyncio's docs. It fails in test B because the usage of the `fixt` fixture overrides the current event loop to be function-scoped, despite the code declares the intention to use the session-scoped event loop.

No warnings are issued, no errors raised, it just silently falls to another event loop — it would be difficult to catch such issues in a big codebase. 

When `fixt` is removed from test B, the sample works. When `fixt` is made session-scoped explicitly, it also works.

Can you please clarify what is the intended outcome in this code below? Should it work?

I also wonder why was it made so that every collector has its own event loop? The most straightforward way at the first glance seems to be to always provide the `event_loop` fixture as before, always function-scoped, which internally either creates a new event loop or escalates to the higher-level `_class_event_loop`, `_module_event_loop`, … `_session_event_loop` — if the test is marked with some higher scope. But maybe I miss some edge-cases or scenarios that prevent such usage?

I would really appreciate it, so that I could adjust my code properly long-term and without hacks.

```python
import asyncio

import pytest
import pytest_asyncio

loop: asyncio.AbstractEventLoop


@pytest_asyncio.fixture  # implies: scope=function    # ❌FAILS
# @pytest_asyncio.fixture(scope='session')              # ✅WORKS
async def fixt() -> None:
    yield


@pytest.mark.asyncio(loop_scope="session")  # uses `_session_event_loop`
async def test_a():
    global loop
    loop = asyncio.get_running_loop()
    print(f"TST {id(asyncio.get_running_loop())=}")


@pytest.mark.asyncio(loop_scope="session")  # uses `_session_event_loop` or `_function_event_loop`
async def test_b(fixt):
    print(f"TST {id(asyncio.get_running_loop())=}")
    assert asyncio.get_running_loop() is loop
```

mathewOracle:
`pytest_configure` warns via a plain `warnings.warn()` when `asyncio_default_fixture_loop_scope` is unset. Per pytest's own docs on `Config.issue_config_time_warning`, warnings raised this way during `pytest_configure` can't be captured the way per-test warnings are -- pytest has no way to install a hookwrapper around `pytest_configure` -- so the warning is silently dropped: it never reaches the warnings summary, and doesn't respect `-W`/`filterwarnings`.

    $ pytest  # asyncio_default_fixture_loop_scope left unset
    2 passed in 0.01s   # no warning shown anywhere, including with -W default

Depending on the caller's global warning filter state it can also raise an uncaught PytestDeprecationWarning straight out of pytest_configure, crashing the run outright (reproducible via pytester's in-process runpytest(), which shares the outer process's filters).

`Config.issue_config_time_warning` is pytest's documented mechanism for exactly this case, and is what pytest's own core plugins use for their config-time deprecations (e.g. _pytest/pastebin.py). Switching to it makes the warning show up in the warnings summary and respect the caller's filters, with no change to the message or when it fires.

Verified live against the exact repro from the issue: the warning now appears in a plain `pytest` run with no flags needed. Added a regression test that fails against the previous code (either silently, or via INTERNALERROR depending on inherited filters) and passes with this change, plus a sanity test that no warning fires when the option is configured. Full test suite shows identical pre-existing failures/errors before and after this change (unrelated compatibility gaps with the installed pytest version); no new failures introduced. ruff check clean.

Closes #1142


pytest_asyncio/plugin.py
```diff
@@ -297,7 +297,16 @@ def pytest_configure(config: Config) -> None:
     default_fixture_loop_scope = config.getini("asyncio_default_fixture_loop_scope")
     _validate_scope(default_fixture_loop_scope, "asyncio_default_fixture_loop_scope")
     if not default_fixture_loop_scope:
-        warnings.warn(PytestDeprecationWarning(_DEFAULT_FIXTURE_LOOP_SCOPE_UNSET))
+        # A plain warnings.warn() here is silently lost: pytest can't wrap
```
tjkuson, line 300: I think we can delete this comment. It reads more like developer commentary pertaining to this PR rather than anything useful long-term.

tests/test_fixture_loop_scopes.py
tjkuson, line 1: Can we simplify these changes? Some of the assertions pass on main and, as-written, obfuscates the behaviour we're trying to protect. I'd also expect a test that covers suppressing this warning through `-W` and `filterwarnings`.

changelog.d/1142.fixed.rst
tjkuson, line 1: Can we make this fragment more precise (a user could mistakenly interpret this as thinking we haven't been emitting this warning at all) and remove the implementation detail about `Config.issue_config_time_warning`?

tjkuson (changes requested):

mathewOracle:
Thanks for the review @tjkuson! Pushed 6014a08 addressing all three points:

- **Comment removed** — agreed, it was PR commentary rather than durable documentation. The rationale now lives in the commit message and the issue.
- **Tests simplified** — dropped the session-scoped fixture/loop machinery that wasn't relevant. There are now two focused tests: one asserting the warning reaches the warnings summary, and a parametrised one covering suppression via both `-W` on the command line and `filterwarnings` in the ini. I checked each assertion fails against `main` (all three fail with `INTERNALERROR ... PytestDeprecationWarning`), so nothing passes for free.
- **Changelog reworded** — drops the `Config.issue_config_time_warning` implementation detail and describes the user-visible behaviour instead, without implying the warning was never emitted.

Note: `tests/test_set_event_loop.py::test_asyncio_run_after_async_fixture_does_not_leak_loop` fails on my machine, but it also fails on `main` (dcd61c2) — unrelated to this change.

PR #1558, opened 2026-09-08
docs: explain loop factory selection for parametrized asyncio tests

lonh-jing:
In strict mode, adding an asyncio marker only through `pytest.param` does not trigger loop-factory parametrization. The parametrization guide currently does not mention this limitation.

Document where to place the marker, with an executable example showing two parameter values running against two loop factories. Also explain why mixed asyncio/Trio parameter sets need separately marked test functions. The example is collected by the existing documentation test suite.

Related to #1463. This documents the existing limitation and a working usage pattern; it does not resolve the underlying runtime issue.

Validation on macOS, Python 3.13.11, pytest 9.1.1:

- The new example passes all four cases with `--asyncio-mode=strict`.
- The parameter-specific marker reproduction fails to select the configured factories in strict mode; the same reproduction selects them in auto mode. Fixture parameter marks were checked separately.
- Sphinx builds with `-n -W --keep-going`; pre-commit checks pass for all changed files.
- Full suite: baseline `294 passed, 3 skipped, 1 failed`; with this patch `298 passed, 3 skipped, 1 failed`. The same existing failure in `test_asyncio_run_after_async_fixture_does_not_leak_loop` raises `PytestAssertRewriteWarning` about `pytest_asyncio` already being imported.

AI assistance: Codex (GPT-6 Astra) assisted with investigation, documentation, and executable examples. The local behavior checks above were actually run.


tjkuson:
Thanks for your interest in improving the documentation!

I'm going to close this because the working usage is already covered by the existing documentation (which this PR somewhat duplicates), and the additional restriction pertains to an unresolved bug. I'd rather keep workaround attached to that issue rather than expand the guide around it. I'll also add that although #1463 is definitely a bug we should solve, the workaround is specific to code example posted by the reporter, and not a pattern that I'd like to bless in the documentation itself.

